PR #1, opened 2023-02-15
Code review

StuartCowley:


src/app.js
```diff
@@ -1,5 +1,199 @@
+/* eslint-disable no-unused-expressions */
```
StuartCowley, line 1: This shouldnt really be needed. The rules are there for a reason, this is getting flagged up because you are using ternary operators to determine what to return in several of the functions. This is easily fixed by the use of an if block (Ill add suggestions to the affected functions for you), so you should delete this line

src/app.js
```diff
@@ -1,5 +1,199 @@
...
+});
+
+app.get('/strings/lower/:string', (req, res) => {
+  res.status(200).json({ result: lowercase(req.params.string) });
+});
+
+app.get('/strings/first-characters/:string', (req, res) => {
+  req.query.length
+    ? res.status(200).json({ result: firstCharacters(req.params.string, req.query.length) })
+    : res.status(200).json({ result: firstCharacter(req.params.string) });
```
StuartCowley, line 51: This is the first of the functions that the eslint issue was flagging up:
```
  if (req.query.length) {
    res.status(200).json({ result: firstCharacters(req.params.string, req.query.length) });
  }
```

src/app.js
```diff
@@ -1,5 +1,199 @@
...
+
+// numbers
+
+app.get('/numbers/add/:num1/and/:num2', (req, res) => {
+  const int1 = parseInt(req.params.num1, 10);
+  const int2 = parseInt(req.params.num2, 10);
+
+  Number.isNaN(int1) || Number.isNaN(int2)
+    ? res.status(400).json({ error: 'Parameters must be valid numbers.' })
+    : res.status(200).json({ result: add(int1, int2) });
```
StuartCowley, line 62: This is the second:
```
  if (Number.isNaN(int1) || Number.isNaN(int2)) {
    res.status(400).json({ error: 'Parameters must be valid numbers.' });
  }
  res.status(200).json({ result: add(int1, int2) });
```

src/app.js
```diff
@@ -1,5 +1,199 @@
...
+});
+
+app.get('/numbers/subtract/:num1/from/:num2', (req, res) => {
+  const int1 = parseInt(req.params.num1, 10);
+  const int2 = parseInt(req.params.num2, 10);
+
+  Number.isNaN(int1) || Number.isNaN(int2)
+    ? res.status(400).json({ error: 'Parameters must be valid numbers.' })
+    : res.status(200).json({ result: subtract(int2, int1) });
```
StuartCowley, line 71: This is the third. I wont add full corrections for the others, but now you see whats causing them it should be an easy thing for you to fix:
```
  if (Number.isNaN(int1) || Number.isNaN(int2)) {
    res.status(400).json({ error: 'Parameters must be valid numbers.' });
  }
  res.status(200).json({ result: subtract(int2, int1) });
```

src/app.js
```diff
@@ -1,5 +1,199 @@
...
+
+  res.status(200).json({ result: remainder(a, b) });
+});
+
+// booleans
+
+app.post('/booleans/negate', (req, res) => {
+  const { value } = req.body;
+
+  res.status(200).json({ result: negate(value) });
```
StuartCowley, line 139: As you are only using the `value` variable once, you dont need to destructure it and create a variable. Can just use it directly in the `res.status` call:
```
```
Theres a few other places where this is done in the rest of the project, generally if you use a value more than once it could be stored in a variable, if not it can be used directly most of the time

src/lib/booleans.js
```diff
@@ -0,0 +1,80 @@
...
+
+function one(a, b) {
+  return (a && !b) || (!a && b);
+}
+
+function truthiness(a) {
+  if (a) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 25: The logic here is fine, but it can be condensed a lot using  two NOT operators (!!) together to create a double negation. What this does is to first evaluate the value to a boolean that will be the opposite of what it actually is, eg passing in a string (which is a truthy value) will be made `false` by one of the ! operators, then the second ! will flip that to `true`, and so evaluating the string as being truthy. Nice trick that gets used more often than you might think!
```
  return !!a
```

StuartCowley (commented):
Good work here, pretty much everything is fixing the code so you dont need to use that eslint exception. Theres some code improvement suggestions too, but nothing major. Also, it would be nice to update the readme as always! 

Nice work :)

PR #2, opened 2023-02-18
Code review

StuartCowley:


src/app.js
```diff
@@ -1,5 +1,190 @@
...
+  if (Number.isNaN(a) || Number.isNaN(b)) {
+    res.status(400).json({ error: 'Parameters must be valid numbers.' });
+  }
+
+  res.status(200).json({ result: remainder(a, b) });
+});
+
+app.post('/booleans/negate', (req, res) => {
+  const { value } = req.body;
+  res.status(200).json({ result: negate(value) });
```
StuartCowley, line 131: Could just as easily put this into one liner, no need to create a variable that only gets used once
```
```

src/app.js
```diff
@@ -1,5 +1,190 @@
...
+});
+
+app.post('/booleans/truthiness', (req, res) => {
+  const { value } = req.body;
+  res.status(200).json({ result: truthiness(value) });
```
StuartCowley, line 136: As above, can be shortened a bit
```
  res.status(200).json({ result: truthiness(req.body.value) });
```

src/app.js
```diff
@@ -1,5 +1,190 @@
...
+
+app.post('/arrays/element-at-index/:index', (req, res) => {
+  const { index } = req.params;
+  const { array } = req.body;
+  res.status(200).json({ result: getNthElement(index, array) });
+});
+
+app.post('/arrays/to-string', (req, res) => {
+  const { array } = req.body;
+  res.status(200).json({ result: arrayToCSVString(array) });
```
StuartCowley, line 168: As above:
```
  res.status(200).json({ result: arrayToCSVString(req.body.array) });
```

src/app.js
```diff
@@ -1,5 +1,190 @@
...
+
+app.post('/arrays/starts-with-vowel', (req, res) => {
+  res.status(200).json({ result: elementsStartingWithAVowel(req.body.array) });
+});
+
+app.post('/arrays/remove-element', (req, res) => {
+  const index = parseInt(req.query.index, 0);
+  console.log(index);
+  const { array } = req.body;
+  if (req.query.index) {
```
StuartCowley, line 183: Couple of things here:
The console statement should be removed, not just for neatness in the code but also as it shows up during the test and clutters the results. Second, why not reuse the `index` variable you created in the opening line of the `if` statement?
```
  const { array } = req.body;
  if (index) {
```

src/lib/arrays.js
```diff
@@ -0,0 +1,118 @@
...
+
+const addToArray2 = (element, array) => {
+  const arrayedTwice = [...array];
+  arrayedTwice.push(element);
+  return arrayedTwice;
+
+  // const contain = [];
+  // contain.push(element);
+  // const newArrayx = array.concat(contain);
+  // return newArrayx;
```
StuartCowley, line 28: Should always remove commented code blocks like this

src/lib/arrays.js
```diff
@@ -0,0 +1,118 @@
...
+};
+
+const onlyEven = numbers => {
+  const hibby = numbers.filter(evenNumber => evenNumber % 2 === 0);
+  return hibby;
+};
+
+// ^^ The above are two examples of returning a function nested in another function.
+//Either with two return statements, or with one, having the nested function
+//assigned to an additional variable.
```
StuartCowley, line 63: Dont really need to include this here

src/lib/arrays.js
```diff
@@ -0,0 +1,118 @@
...
+
+const removeNthElement2 = (index, array) => {
+  return array.filter((_, indexRem) => {
+    return indexRem !== index;
+  });
+};
+
+// const elementsStartingWithAVowel = strings => {
+//   return strings.filter(str => /^[aeiou]/i.test(str));
+// };
```
StuartCowley, line 73: Theres a few more instances of code being commented out like this in the rest of the file, you could remove all of these and would make things a bit clearer

__tests__/numbers.test.js
```diff
@@ -77,18 +77,19 @@ describe('/numbers', () => {
   });
 
   describe('POST /multiply', () => {
-    xit('multiplies two numbers', done => {
+    it('multiplies two numbers', done => {
       request(app)
         .post('/numbers/multiply')
         .send({ a: 10, b: 3 })
         .then(res => {
+          console.log('Response received!');
```
StuartCowley, line 85: Should always remove console statements, especially debugging type ones like this

src/lib/objects.js
```diff
@@ -0,0 +1,74 @@
+/* eslint-disable object-shorthand */
+const createPerson = (name, age) => {
+  return {
+    name: name,
+    // eslint-disable-next-line object-shorthand
+    age: age
+  };
+};
```
StuartCowley, line 8: These eslint exceptions are not needed, you should remove these. Just for reference, eslint is not something you need to fight with as it is trying to help you (though it may not feel like it sometimes). Once you format the file after removing these it will correct these things for you

src/lib/objects.js
```diff
@@ -0,0 +1,74 @@
...
+  return object.name;
+};
+
+const getProperty = (property, object) => {
+  return object[property];
+};
+
+const hasProperty = (property, object) => {
+  // eslint-disable-next-line no-prototype-builtins
+  return object.hasOwnProperty(property);
```
StuartCowley, line 20: As above, you can remove this rule as it isnt needed here. To get around the error, its telling you to use another method to access to objects property, which we can easily do using the square bracket notation instead:
```
  return object[property];
```

StuartCowley (commented):
Nice work! Have left a few suggestions on how to get around the issues youre using eslint exceptions to avoid, as well as a few minor improvements to tidy things up a bit. Nothing major though, all the tests pass and everythings pretty nicely done from the outset. Good stuff

PR #4, opened 2023-02-19
Laura crawford code review

StuartCowley:


README.md
```diff
@@ -1,38 +1 @@
...
-```
-
-## Running tests
-
-You can run tests with the `npm test` command.
-
-## Remember: READMEs are important!
-
-When you're finished, be sure to rewrite this one so it's more specific to your project.
+Used previous JS basic exercises to practice express testing.
```
StuartCowley, line 1: You should always have a heading element (made with a single #) at the start of your readme containing your title. Remember your README is the first thing anybody sees when looking at your repo, so it should have all the information needed to install and use your app as well as an explanation of what it does.

Some things you should include/do:
- what technologies are used,
- installation instructions,
- A description of what the app actually does
- your name!

Also useful is [markdown live preview](https://markdownlivepreview.com/) so you can play with what the file will look like without having to do multiple commits to the repo to see the results, have a look through this [guide to markdown](https://www.markdownguide.org/cheat-sheet/) too

src/app.js
```diff
@@ -1,5 +1,152 @@
 const express = require('express');
+// const bodyParser = require('body-parser');
+
```
StuartCowley, line 3: You shouldnt include commented out code like this, should delete this

src/app.js
```diff
@@ -1,5 +1,152 @@
...
 const app = express();
 
+app.use(express.json());
+
+// STRINGS //
+app.get('/strings/hello/:thing', (req, res) => {
+  const { thing } = req.params;
+  res.status(200);
+  res.send({ result: strings.sayHello(thing) });
+});
```
StuartCowley, line 18: You could condense this a little, also its worth thinking about the names you choose for variable names. 'thing' is a super vague name that doesnt give anything away about what it could actually represent you its always better to use a name that will make it opbvious to whoever is reading this in the future what it is you are actually doing:
```
app.get('/strings/hello/:string', (req, res) => {
  res.status(200);
  res.send({ result: strings.sayHello(req.params.string) });
});
```

src/app.js
```diff
@@ -1,5 +1,152 @@
...
+
+app.get('/numbers/add/:a/and/:b', (req, res) => {
+  const { a } = req.params;
+  const { b } = req.params;
+  if (isNaN(a) || isNaN(b)) {
+    res.status(400);
+    res.send({ error: 'Parameters must be valid numbers.' });
+  }
+  res.status(200);
+  res.send({ result: numbers.add(+a, +b) });
```
StuartCowley, line 50: You could add a clear line between the vairable declaration, function logic, and the send method at the end to make it a bit easier to read at a glance:
```
  const { a } = req.params;
  const { b } = req.params;

  if (isNaN(a) || isNaN(b)) {
    res.status(400);
    res.send({ error: 'Parameters must be valid numbers.' });
  }

  res.status(200);
  res.send({ result: numbers.add(+a, +b) });
```

src/app.js
```diff
@@ -1,5 +1,152 @@
...
+
+app.get('/numbers/subtract/:a/from/:b', (req, res) => {
+  const { a } = req.params;
+  const { b } = req.params;
+  if (isNaN(a) || isNaN(b)) {
+    res.status(400);
+    res.send({ error: 'Parameters must be valid numbers.' });
+  }
+  res.status(200);
+  res.send({ result: numbers.subtract(Number(b), Number(a)) });
```
StuartCowley, line 61: As above:
```
  const { a } = req.params;
  const { b } = req.params;

  if (isNaN(a) || isNaN(b)) {
    res.status(400);
    res.send({ error: 'Parameters must be valid numbers.' });
  }

  res.status(200);
  res.send({ result: numbers.subtract(Number(b), Number(a)) });
```

src/app.js
```diff
@@ -1,5 +1,152 @@
...
+  }
+  res.status(200);
+  res.send({ result: numbers.multiply(Number(b), Number(a)) });
+});
+
+// BOOLEANS //
+app.post('/booleans/negate', (req, res) => {
+  res.status(200);
+  const { value } = req.body;
+  res.send({ result: booleans.negate(value) });
```
StuartCowley, line 82: You dont need to declare a variable to only use it once, could just use the value directly in the res.send statement:
```
  res.send({ result: booleans.negate(req.body.value) });
```

src/app.js
```diff
@@ -1,5 +1,152 @@
...
+});
+
+app.post('/booleans/truthiness', (req, res) => {
+  res.status(200);
+  const { value } = req.body;
+  res.send({ result: booleans.truthiness(value) });
```
StuartCowley, line 88: As above:
```
  res.status(200);
  res.send({ result: booleans.truthiness(req.body.value) });
```

src/app.js
```diff
@@ -1,5 +1,152 @@
...
+
+app.get('/booleans/:string/starts-with/:char', (req, res) => {
+  const { string } = req.params;
+  const { char } = req.params;
+  if (char.length > 1) {
+    res.status(400);
+    res.send({ error: 'Parameter "character" must be a single character.' });
+  }
+  res.status(200);
+  res.send({ result: booleans.startsWith(char, string) });
```
StuartCowley, line 109: As above, could format this a little cleaner:
```
  const { char } = req.params;

  if (char.length > 1) {
    res.status(400);
    res.send({ error: 'Parameter "character" must be a single character.' });
  }

  res.status(200);
  res.send({ result: booleans.startsWith(char, string) });
```

src/app.js
```diff
@@ -1,5 +1,152 @@
...
+});
+
+// ARRAYS //
+
+app.post('/arrays/element-at-index/:index', (req, res) => {
+  res.status(200);
+  const { index } = req.params;
+  const { array } = req.body;
+  // const { index } = req.body;
+  res.send({ result: arrays.getNthElement(index, array) });
```
StuartCowley, line 119: As before, you could condense this a bit
```
  res.status(200);
  res.send({ result: arrays.getNthElement(req.params.index, req.body.array) });
```

src/app.js
```diff
@@ -1,5 +1,152 @@
...
+});
+
+app.post('/arrays/to-string', (req, res) => {
```
StuartCowley, line 122: The rest of the functions here could be reformatted using the same logic as above, would be good to tweak them so its all uniform for the whole file!

src/lib/arrays.js
```diff
@@ -0,0 +1,112 @@
...
+  return newArray;
+};
+
+const removeNthElement = (index, array) => {
+  return array.splice(index, 1);
+};
+
+const numbersToStrings = numbers => {
+  const newString = numbers.toString();
+  return newString.split(',');
```
StuartCowley, line 33: You can chain these methods in a single statement:
```
  return numbers.toString().split(',');
```

src/lib/arrays.js
```diff
@@ -0,0 +1,112 @@
...
+};
+
+const uppercaseWordsInArray = strings => {
+  // return strings.toString().toUpperCase().split(",");
```
StuartCowley, line 37: Should remove commented out code like this

src/lib/arrays.js
```diff
@@ -0,0 +1,112 @@
...
+    return i !== index;
+  });
+};
+
+const elementsStartingWithAVowel = strings => {
+  return strings.filter(string => {
+    const vowels = 'aeiou';
+    const lowerCaseString = string.toLowerCase();
+    const firstLetter = lowerCaseString[0];
+    return vowels.includes(firstLetter);
```
StuartCowley, line 69: You could replace these three lines with just one that uses all the same logic:
```
    return vowels.includes(string.toLowerCase()[0]);
```

src/lib/arrays.js
```diff
@@ -0,0 +1,112 @@
...
+const sortByLastLetter = strings => {
+  const alphabet = 'abcdefghijklmnopqrstuvwxyz';
+  return strings.sort((current, next) => {
+    const currentLastLetterIndex = current.length - 1;
+    const currentLastLetter = current[currentLastLetterIndex];
+    const nextLastLetterIndex = next.length - 1;
+    const nextLastLetter = next[nextLastLetterIndex];
+    const currentIndexInAlphabet = alphabet.indexOf(currentLastLetter);
+    const nextIndexInAlphabet = alphabet.indexOf(nextLastLetter);
+    return currentIndexInAlphabet - nextIndexInAlphabet;
```
StuartCowley, line 92: This is an interesting way to go about this! Theres many different ways you could solve this, so rather than go into them all here Ill just stick to what you could do to refactor this version.
There are a lot of variables created here that are only used once, which is generally bad practice. Looking through your function and putting the variables values in where they are used, it can be boiled down to just this:
```
    return alphabet.indexOf(current[current.length - 1]) - alphabet.indexOf(next[next.length - 1]);
```
Interesting approach as I say, good to see you working these things out for yourself and hitting your won conclusions!

src/lib/booleans.js
```diff
@@ -0,0 +1,131 @@
...
+    return true;
+  }
+  return false;
+}
+
+function either(a, b) {
+  if (a || b === true) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 20: You could replace this with just this, since you are only looking if one of the other is true you can just assert for that:
```
  return a || b;
```

src/lib/booleans.js
```diff
@@ -0,0 +1,131 @@
...
+};
+
+function none(a, b) {
+  if (a === false && b === false){
+    return true;
+  }
+  return false;
```
StuartCowley, line 27: You could simplify this to simply:
```
  return !(a || b);
```

src/lib/booleans.js
```diff
@@ -0,0 +1,131 @@
...
+};
+
+function one(a, b) {
+  if (a === true && b === false) {
+    return true;
+  }
+  if (a === false && b === true) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 37: You could be a bit clever with this one, and refactor a touch:
```
  return (a || b) && !(a && b);
```

src/lib/booleans.js
```diff
@@ -0,0 +1,131 @@
...
+};
+
+function truthiness(a) {
+  if (a) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 44: The logic here is fine, but it can be condensed a lot using two NOT operators (!!) together to create a double negation. What this does is to first evaluate the value to a boolean that will be the opposite of what it actually is, eg passing in a string (which is a truthy value) will be made `false` by one of the ! operators, then the second ! will flip that to `true`, and so evaluating the string as being truthy. Nice trick that gets used more often than you might think!
```
  return !!a
```

src/lib/booleans.js
```diff
@@ -0,0 +1,131 @@
...
+};
+
+function isGreaterThan(a, b) {
+  if (a > b){
+    return true;
+  }
+  return false;
```
StuartCowley, line 58: This can be reduced quite a bit:
```
  return (a > b);
```

src/lib/booleans.js
```diff
@@ -0,0 +1,131 @@
...
+};
+
+function isSquare(a) {
+  if (Math.sqrt(a) % 1 === 0){
+    return true;
+  }
+  return false;
+};
```
StuartCowley, line 87: All these could be reduced in the same way as detailed above. When you are looking to return a true and false value dependent on the evaluation of an expression, you can just return the evaluation of the expression as in the examples above. Makes it much cleaner and easier straight off :)

src/lib/booleans.js
```diff
@@ -0,0 +1,131 @@
...
+function startsWith(char, string) {
+  if (string.startsWith(char)){
+    return true;
+  }
+  return false;
+};
+
+function containsVowels(string) {
+  const vowels = 'aeiou';
+const lowerCaseString = string.toLowerCase();
```
StuartCowley, line 98: Indentation is a bit off here:
```
  const lowerCaseString = string.toLowerCase();
```

src/lib/numbers.js
```diff
@@ -0,0 +1,57 @@
+function add (a, b) {
```
StuartCowley, line 1: The formatting here is a bit out of whack, it would be a great idea to run the formatter across this and let it do the heavy lifting for you

src/lib/objects.js
```diff
@@ -0,0 +1,76 @@
...
+};
+
+const getAges = people => {
+  return people.map(person => {
+    return person.age;
+  });
+};
+
+const findByName = (name, people) => {
+  // your code here
```
StuartCowley, line 32: Should remove the comment here

StuartCowley (commented):
Have left a few comments around formatting and some suggestion around refactoring some of the logic. It would be good to get the missing divide and remainder functions tested too, it would be good to add the fact you intent to fix these to the readme (along with a few other things, have left a comment about the readme on the file itself)

If you havent set up prettier/eslint, Id really recommend doing that as it will save you a lot of work in the future, plus its an industry standard that its great to get familiar with early on

If you have any questions about anything here, feel free to ask :)

PR #5, opened 2023-02-20
Anthony moran code review

StuartCowley:


README.md
```diff
@@ -1,6 +1,6 @@
 # JavaScript Basics in Express
 
-## Getting started
+## Getting started - Remember to personalise for the work I did using the README guides previously posted by other students, esp Cruise Ships!
```
StuartCowley, line 3: Think you may have forgotten to do this :D

src/app.js
```diff
@@ -1,5 +1,206 @@
...
+app.get('/strings/hello/world', (_, res) => {
+  res.status(200).json({ result: 'Hello, world!' });
+});
+
+app.get('/strings/hello/:string', (req, res) => {
+  res.status(200).json({ result: sayHello(req.params.string) });
+});
+
+app.get('/strings/upper/:string', (req, res) => {
+  // const result = req.params.string.toUpperCase();
```
StuartCowley, line 35: Should always remove commented out code like this, it usually just serves to clutter up your work

src/app.js
```diff
@@ -1,5 +1,206 @@
...
+  res.status(200).json({ result: lowercase(req.params.string) });
+});
+
+app.get('/strings/first-character/:string', (req, res) => {
+  res.status(200).json({ result: firstCharacter(req.params.string) });
+});
+
+app.get('/strings/first-characters/:string', (req, res) => {
+  const n = req.query.length;
+  res.status(200).json({ result: firstCharacters(req.params.string, n) });
```
StuartCowley, line 49: No need to create a variable here, just reference the value direct in the return statement:
```
  res.status(200).json({ result: firstCharacters(req.params.string, req.query.length) });
```

src/app.js
```diff
@@ -1,5 +1,206 @@
...
+});
+
+// Numbers Section
+
+app.get('/numbers/add/:a/and/:b', (req, res) => {
+  const a = parseInt(req.params.a, 10);
+  const b = parseInt(req.params.b, 10);
+  if (isNaN(a) || isNaN(b)) {
```
StuartCowley, line 57: Theres an eslint error coming up here, easily fixed by referencing the Number object the method comes from:
```
```

src/app.js
```diff
@@ -1,5 +1,206 @@
...
+    res.status(400).json({ error: 'Parameters must be valid numbers.' });
+  }
+  res.status(200).json({ result: add(a, b) });
+});
+
+app.get('/numbers/subtract/:a/from/:b', (req, res) => {
+  const a = parseInt(req.params.a, 10);
+  const b = parseInt(req.params.b, 10);
+  if (isNaN(a) || isNaN(b)) {
```
StuartCowley, line 66: As above, should update all references to this to use `Number.isNan()` instead

src/app.js
```diff
@@ -1,5 +1,206 @@
...
+    res.status(400).json({ error: 'Parameters must be valid numbers.' });
+  }
+  res.status(200).json({ result: subtract(b, a) });
+  console.log({ result: subtract(b, a) });
```
StuartCowley, line 70: Theres a ton of console logs littered throughout the project, they are really cluttering up the output of the tests! Should definitely remove all of these throughout

src/app.js
```diff
@@ -1,5 +1,206 @@
...
+    res.status(200).json({ result: remainder(req.body.a, req.body.b) });
+  }
+  console.log({ result: remainder(a, b) });
+});
+
+// Booleans Section
+
+app.post('/booleans/negate', (req, res) => {
+  const { value } = req.body;
+  res.status(200).json({ result: negate(value) });
```
StuartCowley, line 142: No need to create a variable here, just reference directly:
```
```
Theres a few instances of this throughout, would be easy enough to update them all to the shorter syntax

src/lib/arrays.js
```diff
@@ -0,0 +1,113 @@
+const getNthElement = (index, array) => {
+  if (index >= array.length) {
+    return array[index - array.length];
+  }
+  return array[index];
```
```
```

src/lib/arrays.js
```diff
@@ -0,0 +1,113 @@
...
+const reverseWordsInArray = strings => {
+  return strings.map(function(A) {
+    return A.split('')
+      .reverse()
+      .join('');
+  });
+};
+
+const onlyEven = numbers => {
+  // const numbers2 = numbers.filter(a => a % 2 === 0); works, replaced filter with map,but this gives T or F//
```
StuartCowley, line 48: Can remove the comment here

src/lib/arrays.js
```diff
@@ -0,0 +1,113 @@
...
+      strings[i].startsWith('A') ||
+      strings[i].startsWith('E') ||
+      strings[i].startsWith('I') ||
+      strings[i].startsWith('O') ||
+      strings[i].startsWith('U')
+    ) {
+      results.push(strings[i]);
+    }
+  }
+  return results;
```
StuartCowley, line 79: The basic idea is here, but theres a few different ways to go about this. This has a lot of repetition, and isnt that maintainable so should be re-written. One way is to define the vowels in a variable, then use a filter to validate if the first letter contains something in that variable eg:

```
  return strings.filter(string => {
    const vowels = 'aeiou';
    return vowels.includes(string.toLowerCase()[0]);
  });
```
Much more concise and does the exact same job

src/lib/booleans.js
```diff
@@ -0,0 +1,88 @@
...
+  return !a && !b;
+}
+
+function one(a, b) {
+  return a !== b;
+}
+
+function truthiness(a) {
+  if (a) return true;
+  return false;
```
StuartCowley, line 23: The logic here is fine, but it can be condensed a lot using two NOT operators (!!) together to create a double negation. What this does is to first evaluate the value to a boolean that will be the opposite of what it actually is, eg passing in a string (which is a truthy value) will be made `false` by one of the ! operators, then the second ! will flip that to `true`, and so evaluating the string as being truthy. Nice trick that gets used more often than you might think!
```
  return !!a
```

src/lib/booleans.js
```diff
@@ -0,0 +1,88 @@
...
+
+function isLessThanOrEqualTo(a, b) {
+  return a <= b;
+}
+
+function isOdd(a) {
+  if (a % 2 !== 0) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 42: You can update a lot of these functions to be more concise by simply returning the evaluation of the condition, for example here:
```
  return (a % 2 !== 0);
```

src/lib/booleans.js
```diff
@@ -0,0 +1,88 @@
...
+}
+
+function startsWith(char, string) {
+  if (char === string.charAt(0)) return true;
+  return false;
+}
+
+function containsVowels(string) {
+  if (string.includes('a') || string.includes('e') || string.includes('O')) return true;
```
StuartCowley, line 63: This could use a bit of an update... see the comment on the method for startsWithAVowel for a cleaner approach

src/lib/objects.js
```diff
@@ -0,0 +1,53 @@
...
+
+const findHondas = cars => cars.filter(object => object.manufacturer === 'Honda')
+
+function averageAge (people) {
+  const totalAge = people.reduce((prevAgeAccumulator, currentValuePerson) => {
+    return prevAgeAccumulator + currentValuePerson.age
+  }, 0)
+  return totalAge / people.length
+}
+// JavaScript Array Reduce - Programming with Mosh, www.youtube.com/watch?v=g1C40tDP0Bk
```
StuartCowley, line 29: Comments like this dont really belong in the repo! Keep notes by all means, but dont put them in the code

src/lib/objects.js
```diff
@@ -0,0 +1,53 @@
...
+const createTalkingPerson = (alias, number) => {
+  return {
+    name: alias,
+    age: number,
+    introduce: strangersName => {
+      return `Hi ${strangersName}, my name is ${alias} and I am ${number}!`
+    }
+  }
+}
+// Use of Template literals (Template strings)
```
StuartCowley, line 40: As above

StuartCowley (commented):
Have left some improvements to the code quality, some refactoring ideas, and a few other things to tidy things up a little in general. When running the tests the whole screen gets filled with console log outputs so definitely get them taken out!

Logic itself is solid all all tests pass, good work!

PR #6, opened 2023-02-22
Tom wilkinson code review

StuartCowley:


src/app.js
```diff
@@ -1,5 +1,199 @@
...
+app.get('/strings/lower/:string', (req, res) => {
+  res.json({ result: lowercase(req.params.string) });
+});
+
+app.get('/strings/first-characters/:string', (req, res) => {
+  const n = req.query.length;
+  if (n) {
+    res.status(200).json({ result: firstCharacters(req.params.string, n) });
+  }
+  res.status(200).json({ result: firstCharacter(req.params.string) });
```
StuartCowley, line 43: You could refactor this a bit. Its a bit of a trivial example but if you are going to use variables for any values that are used more than once in the function you should keep it consistent and do it for all of them. As an example:
```
  const { length } = req.query;
  if (length) {
  }
```

src/app.js
```diff
@@ -1,5 +1,199 @@
...
+  }
+
+  res.status(200).json({ result: remainder(a, b) });
+});
+
+// BOOLEANS
+
+app.post('/booleans/negate', (req, res) => {
+  const { value } = req.body;
+  res.status(200).json({ result: negate(value) });
```
StuartCowley, line 143: You dont need to create a variable here since its value is only being used once:
```
```

src/app.js
```diff
@@ -1,5 +1,199 @@
...
+});
+
+app.post('/booleans/truthiness', (req, res) => {
+  const { value } = req.body;
+  res.status(200).json({ result: truthiness(value) });
```
StuartCowley, line 148: As above, dont need to create the variable here:
```
  res.status(200).json({ result: truthiness(req.body.value) });
```

src/app.js
```diff
@@ -1,5 +1,199 @@
...
+  res.status(200).json({ result: isOdd(number) });
+});
+
+app.get('/booleans/:string/starts-with/:character', (req, res) => {
+  const { string } = req.params;
+  const { character } = req.params;
+  if (character.length > 1) {
+    res.status(400).json({ error: 'Parameter "character" must be a single character.' });
+  }
+  res.status(200).json({ result: startsWith(character, string) });
```
StuartCowley, line 166: As above, you dont need to create a variable for something that only gets used once
```
  const { character } = req.params;
  if (character.length > 1) {
    res.status(400).json({ error: 'Parameter "character" must be a single character.' });
  }
  res.status(200).json({ result: startsWith(character, req.params.string) });
```

src/app.js
```diff
@@ -1,5 +1,199 @@
...
+});
+
+// Arrays
+
+app.post('/arrays/element-at-index/:index', (req, res) => {
+  const { array } = req.body;
+  const { index } = req.params;
+  res.status(200).json({ result: getNthElement(index, array) });
```
StuartCowley, line 174: You could get rid of the variable declarations here:
```
```

src/app.js
```diff
@@ -1,5 +1,199 @@
...
+});
+
+app.post('/arrays/to-string', (req, res) => {
```
StuartCowley, line 177: All the following functions could be re-written in the same way as above

src/lib/arrays.js
```diff
@@ -0,0 +1,95 @@
+const getNthElement = (index, array) => {
+  if (index >= array.length) {
+    return array[index - array.length];
+  }
+  return array[index];
```
```
```

src/lib/arrays.js
```diff
@@ -0,0 +1,95 @@
...
+const sortByLastLetter = strings => {
+  const reverseString = item =>
+    item
+      .split('')
+      .reverse()
+      .join('');
+  return strings
+    .map(reverseString)
+    .sort()
+    .map(reverseString);
```
StuartCowley, line 76: This is an interesting approach! There are multiple ways to do this, another would be using just the sort function:
```
  return strings.sort((a, b) => a.charCodeAt(a.length - 1) - b.charCodeAt(b.length - 1));
```

src/lib/objects.js
```diff
@@ -0,0 +1,70 @@
...
+
+const getProperty = (property, object) => {
+  return object[property];
+};
+
+const hasProperty = (property, object) => {
+  if (object[property]) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 20: When you are looking for return a boolean as you are here, you could use the logic you have here but with a bit of refactoring. If you were to first reduce this to a ternary:
```
```
```
  return !!object[property];
```
Much cleaner :)

src/lib/objects.js
```diff
@@ -0,0 +1,70 @@
...
+};
+
+const isOver65 = person => {
+  if (person.age > 65) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 27: As above, you could simply return the result of evaluating the condition since you are only looking for a boolean value here:
```
  return person.age > 65;
```

StuartCowley (commented):
Good work! Everything passes, have just left some suggestions to refactor some logic to make it a bit more concise but nothing major. Theres some interesting ideas in your approaches here, good stuff to see. Keep it up!

PR #7, opened 2023-02-23
Joseph redmond code review

StuartCowley:


src/app.js
```diff
@@ -2,4 +2,194 @@ const express = require('express');
...
+  res.status(200);
+  res.json({ result: lowercase(req.params.string) });
+});
+
+app.get('/strings/first-characters/:string', (req, res) => {
+  if (Object.keys(req.query).length === 0) {
+    res.status(200).json({ result: firstCharacter(req.params.string) });
+  } else {
+    res.status(200).json({ result: firstCharacters(req.params.string, req.query.length) });
+  }
```
StuartCowley, line 37: There are a few things we can do here to improve this. First, whenever you are using the same value more than once, its generally better to store that value in a variable so it can be re-used without having to add any unnecessary duplication. In this case, `req.params.string` and `req.query.length`.
Then, we can use conditional logic to determine which of the 2 functions to call depending on the value of the `length` property, which will either be `undefined` if not passed in or a number. Putting these together gives us something like:
```
  const { length } = req.query;
  if (!length) {
  } else {
  }
```
Just a bit more concise and clear for the different paths this function could take

Rather than putting the same logic on all the functions that this would apply to, have a read through them all and see what you could refactor in this way. Theres potential to cut a large chunk of the code out of the files here using this approach

src/app.js
```diff
@@ -2,4 +2,194 @@ const express = require('express');
...
+});
+
+app.get('/numbers/add/:num1/and/:num2', (req, res) => {
+  const num1 = parseInt(req.params.num1);
+  const num2 = parseInt(req.params.num2);
```
StuartCowley, line 42: Theres an eslint error here caused by not including the radix property as the second argument of the `parseInt` function. All this refers to is the numerical base system used as the base for the parsing of the int, 10 refers to base 10 (so the normal counting numbers), just as you could use 8 for octal or 2 for binary:
```
  const num1 = parseInt(req.params.num1, 10);
  const num2 = parseInt(req.params.num2, 10);
```

src/app.js
```diff
@@ -2,4 +2,194 @@ const express = require('express');
...
+
+  if (isNaN(num1) || isNaN(num2)) {
```
StuartCowley, line 44: Another eslint fix here. The `isNan` should be written as a property of the `Number` class to provide greater clarity on where the function originates:
```
  if (Number.isNaN(num1) || Number.isNaN(num2)) {
```

It would be an easy fix to go through the project and find other instances of this and update them too!

src/app.js
```diff
@@ -2,4 +2,194 @@ const express = require('express');
...
+
+  if (isNaN(a) || isNaN(b)) {
+    res.status(400).json({ error: 'Parameters must be valid numbers.' });
+  } else {
+    res.status(200).send({ result: remainder(a, b) });
+  }
+});
+
+// HAVE A LOOK AND RE VISIT THE STRINGS
+app.post('/arrays/element-at-index/:index', (req, res) => {
```
StuartCowley, line 115: For all functions you havent finished, a good way to keep track of what each of these should do would be leave a TODO comment on each function detailing what function would be under test (`getNthElement`) in this case. Just makes it easier on you when you come back to this at some point in the future and you cant remember quite what you were trying to do!

src/lib/arrays.js
```diff
@@ -0,0 +1,86 @@
+const getNthElement = (index, array) => {
+  if (array > array.length - 1) {
+    index -= array.length;
+    return array[index];
+  }
+  return array[index];
```
StuartCowley, line 6: You can simplify this quite a bit using the modulus operator, which is similar to division except that instead of returning a decimal when dividing any remainder, it gives a whole number (EG 5 % 3 = 2). 
```
```

src/lib/booleans.js
```diff
@@ -0,0 +1,116 @@
...
+function either(a, b) {
+  return a || b;
+}
+
+function none(a, b) {
+  if (a === false && b === false) {
+    return true;
+  }
+
+  return false;
```
StuartCowley, line 18: You can rewrite this function so it will return false based on one condition, and return true if not:
```
  if (a || b) {
    return false;
  }
  return true;
```
```
```
```
  return !(a || b);
```

src/lib/booleans.js
```diff
@@ -0,0 +1,116 @@
...
+}
+
+function one(a, b) {
+  return Boolean(a ^ b);
+  // NEED TO REVIEW THE ABOVE*//
```
StuartCowley, line 23: You could simply use this:
```
  return Boolean(a);
```

src/lib/booleans.js
```diff
@@ -0,0 +1,116 @@
...
+function truthiness(a) {
+  return Boolean(a);
+}
+
+function isEqual(a, b) {
+  if (a === b) {
+    return true;
+  }
+
+  return false;
```
StuartCowley, line 35: You can update a lot of these functions to be more concise by simply returning the evaluation of the condition, for example here:
```
return (a === b)
```

src/lib/numbers.js
```diff
@@ -0,0 +1,67 @@
+function add (a, b) {
+  const result = a + b;
+  return result;
```
StuartCowley, line 3: You can remove the declaration here and just return the result of the expression:
```
  return a + b;
```
You could also do this for all the other functions here

src/lib/numbers.js
```diff
@@ -0,0 +1,67 @@
...
+}
+
+function quotient (a, b) {
+return parseInt(a/b); 
+
+}
+
+function remainder (a, b) {
+  return parseInt(a%b);
+}
```
StuartCowley, line 53: You should check for formatting by running prettier across your whole project, it will fix things like this indentation and spacing inconsistency here for you:
```
function quotient (a, b) {
  return parseInt(a / b); 
}

function remainder (a, b) {
  return parseInt(a % b);
}
```

src/lib/objects.js
```diff
@@ -0,0 +1,64 @@
+const createPerson = (name, age) => {
+  const person = {
+    name: name,
+    age: age
```
StuartCowley, line 4: Eslint is complaining here as you can use the shorthand form for this:
```
    name,
    age,
```

src/lib/objects.js
```diff
@@ -0,0 +1,64 @@
...
+const getName = object => {
+  return object.name;
+};
+
+const getProperty = (property, object) => {
+  return object[property];
+};
+
+const hasProperty = (property, object) => {
+ return object.hasOwnProperty(property);
```
StuartCowley, line 18: When you are looking for return a boolean as you are here, you could use the logic you have here but with a bit of refactoring. If you were to first reduce this to a ternary:
```
```
```
  return !!object[property];
```
Much cleaner :)

src/lib/objects.js
```diff
@@ -0,0 +1,64 @@
...
+};
+
+const isOver65 = person => {
+  return person.age > 65 ? true:false;
```
StuartCowley, line 22: You dont actually need the whole ternary here:
```
  return person.age > 65;
```

src/lib/objects.js
```diff
@@ -0,0 +1,64 @@
...
+
+const createTalkingPerson = (name, age) => {
+  const person = {
+    name: name,
+    age: age,
+    introduce: (name) => {
+      return `Hi ${name}, my name is ${person.name} and I am ${person.age}!`
+    }
+  };
+  return person;
```
StuartCowley, line 50: Couple of things here, you can use the shorthand for the object properties as in the `createPerson` method above, and eslint is complaining about the use of an identically named variable in the upper scope (in this case `name`) for the placeholder value in the `createTalkingPerson`. Easily fixed by swapping it out for something else:
```
  const person = {
    name,
    age,
    introduce: stranger => {
      return `Hi ${stranger}, my name is ${person.name} and I am ${person.age}!`;
    },
  };
  return person;
```

src/lib/arrays.js
```diff
@@ -0,0 +1,86 @@
...
+  return strings.map(word =>
+    word
+      .split('')
+      .reverse()
+      .join(''),
+  );
+};
+
+const onlyEven = numbers => {
+  return numbers.filter(a => numbers.indexOf(a) % -2 === 1);
```
StuartCowley, line 47: You were so close here! Would be a good idea to read up a bit more about the how the filter method works just so its good and solid for you. In this case you could do this:
```
  return numbers.filter(number => number % 2 === 0);
```

src/lib/objects.js
```diff
@@ -0,0 +1,64 @@
...
+const getAges = people => {
+  return people.map(person => person.age);
+};
+
+const findByName = (name, people) => {
+  return people.find(findPerson => findPerson.name === name);
+};
+
+const findHondas = cars => {
+ return cars.filter(findManufacturer => findManufacturer.manufacturer === 'honda');
```
StuartCowley, line 34: Slight tweak here!
```
```

src/lib/arrays.js
```diff
@@ -0,0 +1,86 @@
...
+};
+
+const removeNthElement2 = (index, array) => {
+  return array.filter(a => a !== array[index]);
```
StuartCowley, line 51: Not quite there with this one:
```
  const newArr = [...array];
  newArr.splice(index, 1);
  return newArr;
```

src/app.js
```diff
@@ -2,4 +2,194 @@ const express = require('express');
...
+app.post('/arrays/starts-with-vowel', (req, res) => {
+  const { array } = req.body;
+  const result = array.filter(str => startsWithVowel(str));
+  res.status(200).json({ result });
+});
+
+app.post('/arrays/remove-element', (req, res) => {
+  const { array } = req.body;
+  const [, ...result] = array;
+  res.status(200).json({ result });
```
StuartCowley, line 154: This needs refactoring to use one of your library methods:
```
  res.status(200).json({ result: removeNthElement2(req.query.index, req.body.array) });
```

StuartCowley (commented):
Have gone over everything here, theres a few suggestions added to improve general readability of your work, some refactors to make things a bit cleaner, and a few other general points to tidy things up a bit

Theres a few of the functions that arent being used in the express side of things, which is a bit of a shame. I unskipped all your tests to see what was causing the problems and have gone through and left suggestions to fix all the broken things, so implementing all these things will at least give you a working suite!
I would recommend having a read through of some of the fundamental array methods like filter too. You will use these things frequently on the job so its worth spending a bit of time to get fully into how they work and properly understand how they can be used, definitely worth your time!

Thanks for submitting this, keep going youve got this

PR #8, opened 2023-02-24
Tarndeep virdi code review

StuartCowley:


README.md
```diff
@@ -1,38 +1,111 @@
...
-    ├── arrays.js
-    ├── booleans.js
-    ├── numbers.js
-    ├── objects.js
-    └── strings.js
-```
+The API contains the following endpoints:
 
-## Running tests
+`/arrays/remove-element`
```
StuartCowley, line 23: Instead of creating a long list all the same format like you have here, why not use a table instead? For example:

```
## Endpoints
### Strings

| HTTP method | Method      | Endpoint | Description |
| ----------- | ----------- |----------- | ----- |
| GET         | sayHello(string) | /strings/hello/:string | Returns the string "Hello, " concatenated with the given string. |
| GET         | uppercase(string) | /strings/upper/:string | Returns the given string converted to all uppercase. |
```
You could create a new table for each grouping like this, adding whatever info you like. Have a look at [this guide](https://www.markdownguide.org/extended-syntax/) for more info on the formatting options possible

README.md
```diff
@@ -1,38 +1,111 @@
...
-## Remember: READMEs are important!
+`/strings/lower/:lower`
+- Method: GET
+- Parameters: lower (request params)
+- Description: Converts the input string to lowercase and returns it in the response.
 
-When you're finished, be sure to rewrite this one so it's more specific to your project.
+`/strings/first-characters/:firstp`
+- Method: GET
+- Parameters: firstp (request params), length (request query)
```
StuartCowley, line 110: Typo in first?
```
- Parameters: first (request params), length (request query)
```

README.md
```diff
@@ -1,38 +1,111 @@
 # JavaScript Basics in Express
 
-## Getting started
+A simple Express API that contains a set of endpoints for manipulating arrays, booleans, and numbers.
 
-Ensure your Visual Studio Code application is up to date, and that you have the [ESLint extension](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) installed and enabled.
+## Requirements
```
StuartCowley, line 5: Lots of detail here, great!

src/app.js
```diff
@@ -2,4 +2,163 @@ const express = require('express');
...
+const { elementsStartingWithAVowel, removeNthElement2 } = require('./lib/arrays');
+
+app.use(express.json());
+
+app.post('/arrays/remove-element', (req, res) => {
+  const { index } = req.query;
+  const { array } = req.body;
+  const indexNum = parseInt(index, 10);
+  const newArr = removeNthElement2(indexNum, array);
+  res.status(200).send({ result: newArr });
```
StuartCowley, line 17: This could be refactored a bit. You dont need to destructure the index from `req.query` to then later pass that value into a `parseInt` function on a new line, could just as easily do it all on one line without sacrificing readability:
```
  const indexNum = parseInt(req.query.index, 10);
  const newArr = removeNthElement2(indexNum, req.body.array);
  res.status(200).send({ result: newArr });
```

src/app.js
```diff
@@ -2,4 +2,163 @@ const express = require('express');
...
+});
+
+app.post('/arrays/starts-with-vowel', (req, res) => {
+  const { array } = req.body;
+  const newArr = elementsStartingWithAVowel(array);
+  res.status(200).send({ result: newArr });
```
StuartCowley, line 23: This could be reduced down to a single line and still be clear:
```
  res.status(200).send({ result: elementsStartingWithAVowel(req.body.array) });
```

src/app.js
```diff
@@ -2,4 +2,163 @@ const express = require('express');
...
+});
+
+app.post('/arrays/to-string', (req, res) => {
+  const { array } = req.body;
+  res.status(200).send({ result: array.join(',') });
```
StuartCowley, line 28: As above:
```
  res.status(200).send({ result: req.body.array.join(',') });
```

src/app.js
```diff
@@ -2,4 +2,163 @@ const express = require('express');
...
+});
+
+app.post('/arrays/element-at-index/:index', (req, res) => {
+  const { index } = req.params;
+  const { array } = req.body;
+  res.status(200).send({ result: array[index] });
```
StuartCowley, line 34: You could get rid of the variable declarations here:
```
```

src/app.js
```diff
@@ -2,4 +2,163 @@ const express = require('express');
...
+});
+
+app.post('/arrays/append', (req, res) => {
+  const { value } = req.body;
+  const [...array] = req.body.array;
+  array.push(value);
+  res.status(200).send({ result: array });
```
StuartCowley, line 41: Slight reduction here:
```
  const [...array] = req.body.array;
  array.push(req.body.value);
  res.status(200).send({ result: array });
```

src/app.js
```diff
@@ -2,4 +2,163 @@ const express = require('express');
...
+  if (char.length > 1) {
+    res.status(400).send({ error: 'Parameter "character" must be a single character.' });
+  } else {
+    res.status(200).send({ result: startsWith(char, string) });
+  }
+});
+
+app.get('/booleans/is-odd/:num', (req, res) => {
+  const { num } = req.params;
+  if (!/^-?\d*\.?\d+$/.test(num)) {
```
StuartCowley, line 55: Nice to see regex experimentation here! This is one of those things that will get easier with time so dont worry if you dont get it right off the bat. In this case, the expression is way more complex than it needs to be. Youre checking that the first character is not a dash/question mark/a digit following by any other character, a full stop, or optionally a digit of any length (I think), when in reality all you need to do is check if it is not a number. You can do this with just
```
if (!/\d/.test(num)) {
```

src/app.js
```diff
@@ -2,4 +2,163 @@ const express = require('express');
...
+    res.status(400).send({ error: 'Parameter must be a number.' });
+  } else {
+    res.status(200).send({ result: isOdd(parseInt(num, 10)) });
+  }
+});
+
+app.post('/booleans/truthiness', (req, res) => {
+  const { value } = req.body;
+  if (truthiness(value) === true) {
```
StuartCowley, line 64: Dont need to declare the variable here:
```
  if (truthiness(req.body.value) === true) {
```

src/app.js
```diff
@@ -2,4 +2,163 @@ const express = require('express');
...
+    res.status(200).send({ result: true });
+  } else {
+    res.status(200).send({ result: false });
+  }
+});
+
+app.post('/booleans/negate', (req, res) => {
+  const { value } = req.body;
+  const x = Boolean(value);
+  res.status(200).send({ result: !x });
```
StuartCowley, line 74: Remember the idea here was to use the function you created in the original JS basics exercise, you should re-write this to use a call to your `negate` function

src/app.js
```diff
@@ -2,4 +2,163 @@ const express = require('express');
...
+});
+
+app.get('/numbers/add/:a/and/:b', (req, res) => {
+  const { a, b } = req.params;
+  if (/^-?\d*\.?\d+$/.test(a) && /^-?\d*\.?\d+$/.test(b)) {
```
StuartCowley, line 79: See comment above about complex regexs! All other instances of this should be looked at throughout

src/app.js
```diff
@@ -2,4 +2,163 @@ const express = require('express');
...
+    res.status(400).send({ error: 'Unable to divide by 0.' });
+  } else if (a == 0) {
+    res.status(200).send({ result: divide(a, b) });
+  } else if (!a || !b) {
+    res.status(400).send({ error: 'Parameters "a" and "b" are required.' });
+  } else if (!/^-?\d*\.?\d+$/.test(a) || !/^-?\d*\.?\d+$/.test(b)) {
+    res.status(400).send({ error: 'Parameters "a" and "b" must be valid numbers.' });
+  } else {
+    res.status(200).send({ result: divide(a, b) });
+  }
```
StuartCowley, line 120: This is a little long winded. If you look carefully you'll see you are making identical calls on lines 113 and 119. You should never need to do this, ideally the conditional logic will aways direct to the correct call avoiding the need for repetition. In this case, you can fix this by tweaking the 2nd else if statement to explicitly check for `undefined` for the 2 input parameters. Once youve done this you can delete the `a === 0` condition:
```
  const { a, b } = req.body;
  if (b === 0) {
  } else if (a === undefined || b === undefined) {
    res.status(400).send({ error: 'Parameters "a" and "b" are required.' });
  } else if (!/^-?\d*\.?\d+$/.test(a) || !/^-?\d*\.?\d+$/.test(b)) {
    res.status(400).send({ error: 'Parameters "a" and "b" must be valid numbers.' });
  } else {
    res.status(200).send({ result: divide(a, b) });
  }
```
Also, its best practice to always use the strict equality operator `===` whenever possible, as it makes your work more robust and less prone to errors

src/app.js
```diff
@@ -2,4 +2,163 @@ const express = require('express');
...
+});
+
+app.post('/numbers/remainder', (req, res) => {
```
StuartCowley, line 123: See comment in above function!

src/app.js
```diff
@@ -2,4 +2,163 @@ const express = require('express');
...
+  } else if (!a || !b) {
+    res.status(400).send({ error: 'Parameters "a" and "b" are required.' });
+  } else if (!/^-?\d*\.?\d+$/.test(a) || !/^-?\d*\.?\d+$/.test(b)) {
+    res.status(400).send({ error: 'Parameters must be valid numbers.' });
+  } else {
+    res.status(200).send({ result: remainder(a, b) });
+  }
+});
+
+app.get('/strings/hello/:string', (req, res) => {
```
StuartCowley, line 138: Remember to use your own functions and not just re-write them here!

src/app.js
```diff
@@ -2,4 +2,163 @@ const express = require('express');
...
+});
+
+app.get('/strings/first-characters/:firstp', (req, res) => {
+  const { firstp } = req.params;
+  const n = parseInt(req.query.length, 10);
+  if (!req.query.length) {
+    res.status(200).send({ result: firstCharacter(firstp) });
+  } else {
+    const x = firstCharacters(firstp, n);
+    res.status(200).send({ result: x });
```
StuartCowley, line 160: Dont need to create the variable here. Also, its worth thinking about what your variables names are when you do need them, something like `x` really isnt clear what it represents
```
    res.status(200).send({ result: firstCharacters(firstp, parseInt(req.query.length, 10)) });
```

src/lib/arrays.js
```diff
@@ -0,0 +1,111 @@
+const getNthElement = (index, array) => {
```
StuartCowley, line 1: You should remove all the comments in this file, they are just adding noise and dont help with anything really

src/lib/arrays.js
```diff
@@ -0,0 +1,111 @@
...
+const removeNthElement = (index, array) => {
+  return array.splice(index, 1);
+};
+
+const numbersToStrings = numbers => {
+  // return numbers.map(String);
+  const numString = numbers.map(number => {
+    return String(number);
+  });
+  return numString;
```
StuartCowley, line 31: No variable creation needed here, can just return the function result directly:
```
  return numbers.map(number => {
    return String(number);
  });
```

src/lib/arrays.js
```diff
@@ -0,0 +1,111 @@
...
+};
+
+const uppercaseWordsInArray = strings => {
+  const newArray = strings.map(string => {
+    return string.toUpperCase();
+  });
+  return newArray;
```
StuartCowley, line 38: As above:
```
  return strings.map(string => {
    return string.toUpperCase();
  });
```

src/lib/arrays.js
```diff
@@ -0,0 +1,111 @@
...
+  //     .join('')
+  // );
+
+  const reverseString = strings.map(string => {
+    return string
+      .split('')
+      .reverse()
+      .join('');
+  });
+  return reverseString;
```
StuartCowley, line 55: As above:
```
  return strings.map(string => {
    return string
      .split('')
      .reverse()
      .join('');
  });
```

src/lib/arrays.js
```diff
@@ -0,0 +1,111 @@
...
+};
+
+const onlyEven = numbers => {
+  const evenOnly = numbers.filter(number => {
+    if (number % 2 === 0) {
+      return number;
+    }
+  });
+  return evenOnly;
```
StuartCowley, line 64: As above
```
  return numbers.filter(number => {
    if (number % 2 === 0) {
      return number;
    }
  });
```

src/lib/arrays.js
```diff
@@ -0,0 +1,111 @@
...
+    const testString = string.toLowerCase();
+    return (
+      testString[0] === 'a' ||
+      testString[0] === 'e' ||
+      testString[0] === 'i' ||
+      testString[0] === 'o' ||
+      testString[0] === 'u'
+    );
+  });
+  return stringNew;
```
StuartCowley, line 84: The basic idea is solid here, but we can definitely re-write this a bit to make it more concise using the `.test()` method and not using an array for the vowels:
```
  return strings.filter(string => /^[aeiou]/i.test(string));
```
Fundamentally the same logic as you are already using, just condensed a bit

src/lib/arrays.js
```diff
@@ -0,0 +1,111 @@
...
+};
+
+const removeSpaces = string => {};
+
+const sumNumbers = numbers => {
+  const sum = numbers.reduce(number => {});
+};
+
+const sortByLastLetter = strings => {};
+
```
StuartCowley, line 94: It would be nice to see these all completed!

src/lib/booleans.js
```diff
@@ -0,0 +1,91 @@
+const { arrayToCSVString } = require('./arrays');
```
StuartCowley, line 1: It would be good to see these completed and added into the test suites. In the meantime, I would add a comment about your intention to complete these in the future to the README

src/lib/numbers.js
```diff
@@ -0,0 +1,63 @@
...
+function subtract(a, b) {
+  return a - b;
+}
+
+function multiply(a, b) {
+  return a * b;
+}
+
+function divide(a, b) {
+  if (a == 0) {
```
StuartCowley, line 14: Nice error catch, though you should use the === strict equality operator here

src/lib/numbers.js
```diff
@@ -0,0 +1,63 @@
...
+    return 0;
+  }
+  return a / b;
+}
+
+function power(a, b) {
+  // eslint-disable-next-line no-restricted-properties
+  return Math.pow(a, b);
```
StuartCowley, line 22: No ned for the eslint exception here, when hovering over the error it even gives you the answer!
```
  return a ** b;
```

src/lib/numbers.js
```diff
@@ -0,0 +1,63 @@
...
+  return Math.floor(a);
+}
+
+function absolute(a) {
+  return Math.abs(a);
+}
+
+function quotient(a, b) {
+  const r = a / b;
+  return Math.trunc(r);
```
StuartCowley, line 43:
```
  return Math.trunc(a / b);
```

src/lib/numbers.js
```diff
@@ -0,0 +1,63 @@
...
+}
+
+function remainder(a, b) {
+  const r = a % b;
+  return r;
```
StuartCowley, line 48:
```
  return a % b;
```

src/lib/objects.js
```diff
@@ -0,0 +1,52 @@
+const createPerson = (name, age) => {
```
StuartCowley, line 1: As above, should adad a note to the readme about your intention to complete these later

src/lib/strings.js
```diff
@@ -0,0 +1,38 @@
...
+function uppercase(string) {
+  return string.toUpperCase();
+}
+
+function lowercase(string) {
+  return string.toLowerCase();
+}
+
+function countCharacters(string) {
+  // your code here
```
StuartCowley, line 14: Should remove all the comments here

StuartCowley (commented):
Some good work here, the readme is pretty extensive, have left a suggestion on how to make it a bit more presentable. You should also add a todo section explaining which parts of the code are still works in progress, again have left comments about which parts I mean. 
Most of the suggestions are around refactoring existing work and making it a bit tidier, while it does all work there are still some things that could be done to make it a bit more concise. Nice to see regex getting a look in here too!

PR #9, opened 2023-02-24
Matt giles code review

StuartCowley:


__tests__/booleans.test.js
```diff
@@ -145,5 +145,18 @@ describe('/booleans', () => {
           done();
         });
     });
+    // Extra test to test .match(/[aeiou]/i) on containsVowels
```
StuartCowley, line 148: Dont need to include the comment here, can just delete this

src/app.js
```diff
@@ -1,5 +1,155 @@
...
+app.get('/strings/lower/:str', (req, res) => {
+  const string = req.params.str;
+  res.status(200).json({ result: lowercase(string) });
+});
+
+app.get('/strings/first-characters/:str', (req, res) => {
+  const string = req.params.str;
+  const i = req.query.length;
+  const n = i > 0 ? i : 1;
+  res.status(200).json({ result: firstCharacters(string, n) });
```
StuartCowley, line 42: Id say you could condense this down a fair bit:
```
  const { length } = req.query;
  res.status(200).json({ result: firstCharacters(req.params.str, length > 0 ? length : 1) });
```

src/app.js
```diff
@@ -1,5 +1,155 @@
...
+});
+
+// Numbers
+
+app.get('/numbers/add/:num1/and/:num2', (req, res) => {
+  const number1 = parseInt(req.params.num1);
+  const number2 = parseInt(req.params.num2);
```
StuartCowley, line 49: Theres an eslint error here caused by not including the radix property as the second argument of the `parseInt` function. All this refers to is the numerical base system used as the base for the parsing of the int, 10 refers to base 10 (so the normal counting numbers), just as you could use 8 for octal or 2 for binary:
```
  const number1 = parseInt(req.params.num1, 10);
  const number2 = parseInt(req.params.num2, 10);
```

src/app.js
```diff
@@ -1,5 +1,155 @@
...
+app.post('/numbers/multiply', (req, res) => {
+  if (typeof req.body.a === 'undefined' || typeof req.body.b === 'undefined') {
+    res.status(400).json({ error: 'Parameters "a" and "b" are required.' });
+  } else {
+    const answer =
+      isNaN(req.body.a) || isNaN(req.body.b)
+        ? res.status(400).json({ error: 'Parameters "a" and "b" must be valid numbers.' })
+        : res.status(200).json({ result: multiply(req.body.a, req.body.b) });
+    return answer;
+  }
```
StuartCowley, line 72: Okay, so theres a few things to look at here. While this works, it uses the global `isNaN()` method which is being replaced by the more robust `Number.isNan` (see [MDN docs on it here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN))

When replacing the `isNaN` function with `Number.isNaN()`, this breaks. This is because you are only checking for whether or not the values passed in are undefined or not, then only whether they are numbers or not. When passing in strings this will break it, so an extra check has to be added in to convert the values to numbers if they are strings using `parseInt()`. 
Since `parseInt` will return `NaN` if passed a string, we can then check use `Number.isNan()` to check if it is actually a number passed in in the first place. Putting all that together, you could do something like this:
```
  let { a, b } = req.body;

  if (a === undefined || b === undefined) {
    res.status(400).json({ error: 'Parameters "a" and "b" are required.' });
  }

  if (typeof a === 'string' || typeof b === 'string') {
    a = parseInt(a, 10);
    b = parseInt(b, 10);
  }

    res.status(400).json({ error: 'Parameters "a" and "b" must be valid numbers.' });
  }

  res.status(200).json({ result: multiply(a, b) });
```

src/app.js
```diff
@@ -1,5 +1,155 @@
...
+app.post('/numbers/remainder', (req, res) => {
+  if (typeof req.body.a === 'undefined' || typeof req.body.b === 'undefined') {
+    res.status(400).json({ error: 'Parameters "a" and "b" are required.' });
+  } else if (isNaN(req.body.a) || isNaN(req.body.b)) {
+    res.status(400).json({ error: 'Parameters must be valid numbers.' });
+  } else if (req.body.b === 0) {
+    res.status(400).json({ error: 'Unable to divide by 0.' });
+  } else {
+    res.status(200).json({ result: remainder(req.body.a, req.body.b) });
+  }
```
StuartCowley, line 96: These tow could be refactored using the ideas from multiply to make them a bit more robust

src/lib/arrays.js
```diff
@@ -0,0 +1,89 @@
+const getNthElement = (index, array) => {
```
StuartCowley, line 1: Im going to go over the `/lib` files as have already reviewed them last time round!

StuartCowley (commented):
Solid stuff here mate, got some unusual ways of looking at these. Have left some improvements around making things a bit more robust, and some minor formatting type things here and there too. 
Make sure you update the README too!

PR #10, opened 2023-02-25
Ernesto Angustia Code Review

StuartCowley:


src/lib/numbers.js
```diff
@@ -0,0 +1,60 @@
...
+function multiply(a, b) {
+  return a * b;
+}
+
+function divide(a, b) {
+  return a / b;
+}
+
+function power(a, b) {
+  return Math.pow(a, b);
```
StuartCowley, line 18: You have an eslint error on this line. Remember if you hover over the code with red lines below it VSCode shows some info on what the issue is and how to resolve it, in this case it tells you to use the exponentiation operator instead:
```
  return a ** b;
```

src/lib/numbers.js
```diff
@@ -0,0 +1,60 @@
...
+
+function quotient(a, b) {
+  return Math.trunc(a / b);
+}
+
+function remainder(a, b) {
+  const c = Math.trunc(a / b);
+  const d = b * c;
+  const e = a - d;
+  return e;
```
StuartCowley, line 45: This works, but is quite long winded for what you are trying to achieve. You have a built in method in JS that will do this for you: the modulus operator:
```
  return a % b;
```

src/app.js
```diff
@@ -1,5 +1,163 @@
...
+app.get('/strings/upper/:string', (req, res) => {
+  res.status(200).json({ result: uppercase(req.params.string) });
+});
+
+app.get('/strings/lower/:string', (req, res) => {
+  res.status(200).json({ result: lowercase(req.params.string) });
+});
+// app.get('/strings/first-characters/:string', (req, res) => {
+// res.status(200).json({ result: firstCharacter(req.params.string) });
+// });
```
StuartCowley, line 35: Commented out code like this should always be removed, you already have a working function covering this so it can be deleted

src/app.js
```diff
@@ -1,5 +1,163 @@
...
+app.get('/strings/first-characters/:string', (req, res) => {
+  let length = 0;
+  if (req.query.length) {
+    length = req.query.length;
+  } else {
+    length = 1;
+  }
+  res.status(200).json({ result: firstCharacters(req.params.string, length) });
+});
```
StuartCowley, line 44: Bit of a suggestion here, as you are crowbarring the functionality of two functions (`firstCharacter` and `firstCharacters`) into one, when you should really make use of both of them. We can do this by using conditional logic to determine which of the 2 functions to call depending on the value of the `length` property, which will either be `undefined` if not passed in or a number. So something like:

```
  const { length } = req.query;
  if (!length) {
  } else {
  }
});
```

src/app.js
```diff
@@ -1,5 +1,163 @@
...
+app.post('/numbers/multiply', (req, res) => {
+  const { a, b } = req.body;
+  if (!a || !b) {
+    res.status(400).json({ error: 'Parameters "a" and "b" are required.' });
+    // eslint-disable-next-line no-restricted-globals
+  } else if (isNaN(a) && isNaN(b)) {
+    res.status(400).json({ error: 'Parameters "a" and "b" must be valid numbers.' });
+  }
+  res.status(200).send({ result: multiply(a, b) });
```
StuartCowley, line 70: Okay, so theres a few things to look at here. While this works, it uses the global `isNaN()` method which is being replaced by the more robust `Number.isNan` (see [MDN docs on it here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN))

When replacing the `isNaN` function with `Number.isNaN()`, this breaks. This is because you are only checking for whether or not the values passed in are undefined or not, then only whether they are numbers or not. When passing in strings this will break it, so an extra check has to be added in to convert the values to numbers if they are strings using `parseInt()`. 
Since `parseInt` will return `NaN` if passed a string, we can then check use `Number.isNan()` to check if it is actually a number passed in in the first place. Putting all that together, you could do something like this:
```
  let { a, b } = req.body;

  if (a === undefined || b === undefined) {
    res.status(400).json({ error: 'Parameters "a" and "b" are required.' });
  }

  if (typeof a === 'string' || typeof b === 'string') {
    a = parseInt(a, 10);
    b = parseInt(b, 10);
  }

    res.status(400).json({ error: 'Parameters "a" and "b" must be valid numbers.' });
  }

  res.status(200).json({ result: multiply(a, b) });
```
You should never use eslint exceptions unless they are absolutely necessary. In this case the code just needs a bit of a rewrite to fix the issue

src/app.js
```diff
@@ -1,5 +1,163 @@
...
+//  ******** Numbers Test *********//
+
+app.get('/numbers/add/:a/and/:b', (req, res) => {
+  const a = parseInt(req.params.a, 10);
+  const b = parseInt(req.params.b, 10);
+  return Number.isNaN(a) || Number.isNaN(b)
+    ? res.status(400).json({ error: 'Parameters must be valid numbers.' })
+    : res.status(200).json({ result: add(a, b) });
+});
+app.get('/numbers/subtract/:a/from/:b', (req, res) => {
```
StuartCowley, line 55: Remember to always leave a gap between each of your functions, it just makes it easier to read and helps break up the file a bit

src/app.js
```diff
@@ -1,5 +1,163 @@
...
+    res.status(200).send({ result: 0 });
+  } else if (b === 0) {
+    res.status(400).send({ error: 'Unable to divide by 0.' });
+  } else if (!a || !b) {
+    res.status(400).json({ error: 'Parameters "a" and "b" are required.' });
+    // eslint-disable-next-line no-restricted-globals
+  } else if (isNaN(a) && isNaN(b)) {
+    res.status(400).json({ error: 'Parameters "a" and "b" must be valid numbers.' });
+  }
+  res.status(200).send({ result: divide(a, b) });
```
StuartCowley, line 84: As above, this should be rewritten to remove the need for the eslint exception:
```
  let { a, b } = req.body;

  if (a === undefined || b === undefined) {
    res.status(400).json({ error: 'Parameters "a" and "b" are required.' });
  }

  if (typeof a === 'string' || typeof b === 'string') {
    a = parseInt(a, 10);
    b = parseInt(b, 10);
  }

    res.status(400).json({ error: 'Parameters "a" and "b" must be valid numbers.' });
  }

  if (b === 0) {
  }

  res.status(200).json({ result: divide(a, b) });
```

src/app.js
```diff
@@ -1,5 +1,163 @@
...
+});
+app.post('/numbers/divide', (req, res) => {
+  const a = parseInt(req.params.b, 10);
+  const b = parseInt(req.params.a, 10);
+  res.status(200).send({ result: divide(a, b) });
+});
```
StuartCowley, line 90: This version does not work, it should be deleted

src/app.js
```diff
@@ -1,5 +1,163 @@
...
+app.post('/numbers/remainder', (req, res) => {
```
StuartCowley, line 91: This should be rewritten as described above for the same reason

src/app.js
```diff
@@ -1,5 +1,163 @@
...
+  }
+  res.status(200).send({ result: remainder(a, b) });
+});
+
+// ********* Arrays Test ********* //
+
+app.post('/arrays/element-at-index/:index', (req, res) => {
+  const { index } = req.params;
+  const { array } = req.body;
+  res.status(200).json({ result: getNthElement(index, array) });
```
StuartCowley, line 111: You could get rid of the variable declarations here:
```
```

src/app.js
```diff
@@ -1,5 +1,163 @@
...
+});
+app.post('/arrays/to-string', (req, res) => {
+  const { array } = req.body;
+  res.status(200).json({ result: arrayToCSVString(array) });
```
StuartCowley, line 115: As above:
```
  res.status(200).json({ result: arrayToCSVString(req.body.array) });
```

src/app.js
```diff
@@ -1,5 +1,163 @@
...
+});
+app.post('/arrays/append', (req, res) => {
+  const { array, value } = req.body;
+  res.status(200).json({ result: addToArray2(value, array) });
```
StuartCowley, line 119: As above this could be refactored a bit. Remember you dont need to create variables to help values that are only used once in the rest of the function
```
  res.status(200).json({ result: addToArray2(req.body.value, req.body.array) });
```

It would be good to go through the rest of the functions here to remove the unneeded variables, nice quick job

src/app.js
```diff
@@ -1,5 +1,163 @@
...
+  res.status(200).json({ result: elementsStartingWithAVowel(array) });
+});
+app.post('/arrays/remove-element', (req, res) => {
+  const { array } = req.body;
+  const index = parseInt(req.query.index, 10);
+  if (req.query.index) {
+    res.status(200).json({ result: removeNthElement2(index, array) });
+  } else {
+    res.status(200).json({ result: removeNthElement2(0, array) });
+  }
```
StuartCowley, line 132: Theres a bit of unnecessary repetition here, you could update this by changing how you handle the index parameter. If its falsy, you want the index value to be 0, otherwise you want to parse it into an integer and use that. So this function can be re-written to just:

```
  const index = req.query.index ? parseInt(req.query.index, 10) : 0;
  res.status(200).json({ result: removeNthElement2(index, req.body.array) });
```

src/app.js
```diff
@@ -1,5 +1,163 @@
...
+  res.status(200).json({ result: negate(value) });
+});
+app.post('/booleans/truthiness', (req, res) => {
+  const { value } = req.body;
+  res.status(200).json({ result: truthiness(value) });
+});
+app.get('/booleans/is-odd/:number', (req, res) => {
+  const value = parseInt(req.params.number, 10);
+  // eslint-disable-next-line no-restricted-globals
+  if (isNaN(value)) {
```
StuartCowley, line 146: You caan remove the exception need here by using the `Number.isNaN` method:
```
  if (Number.isNaN(value)) {
```

src/app.js
```diff
@@ -1,5 +1,163 @@
...
+});
+app.get('/booleans/:word/starts-with/:character', (req, res) => {
+  const { character } = req.params;
+  const { word } = req.params;
+  // eslint-disable-next-line no-restricted-globals
+  if (character.length > 1) {
+    res.status(400).json({ error: 'Parameter "character" must be a single character.' });
+  } else {
+    res.status(200).json({ result: startsWith(character, word) });
+  }
```
StuartCowley, line 160: Small update here to make it a bit shorter and to remove the eslint rule you dont need:
```
  const { character } = req.params;
  if (character.length > 1) {
    res.status(400).json({ error: 'Parameter "character" must be a single character.' });
  } else {
    res.status(200).json({ result: startsWith(character, req.params.word) });
  }
```

src/lib/arrays.js
```diff
@@ -0,0 +1,104 @@
...
+  return string.split(',', 5);
+};
+
+const addToArray = (element, array) => {
+  array.push(element);
+};
+
+const addToArray2 = (element, array) => {
+  const newArray = array.concat(element);
+  return newArray;
```
StuartCowley, line 19: You can remove the unneeded variable creation here:
```
  return array.concat(element);
```

src/lib/arrays.js
```diff
@@ -0,0 +1,104 @@
...
+
+const removeNthElement = (index, array) => {
+  return array.splice(index, 1);
+};
+
+const numbersToStrings = numbers => {
+  const numtoString = numbers.map(num => {
+    return String(num);
+  });
+  return numtoString;
```
StuartCowley, line 30: Theres no need to create the function here. You can just return the value of the map function directly without the additional call:
```
  return numbers.map(num => {
    return String(num);
  });
```

src/lib/arrays.js
```diff
@@ -0,0 +1,104 @@
...
+};
+
+const uppercaseWordsInArray = strings => {
+  function capitalizeWords(strings) {
+    return strings.map(element => {
+      return element.toUpperCase();
+    });
+  }
+  return capitalizeWords(strings);
```
StuartCowley, line 39: Again, no need for the extra function here:
```
  return strings.map(element => {
    return element.toUpperCase();
  });
```

src/lib/arrays.js
```diff
@@ -0,0 +1,104 @@
...
+};
+
+const reverseWordsInArray = strings => {
+  return strings.map(function(A) {
+    return A.split('')
+      .reverse()
+      .join('');
+  });
```
StuartCowley, line 47: Convention with variable names is to make them something meaningful. In this case, `A` is very abstract and should be replaced with something like the singular name for the array being mapped over:
```
  return strings.map(string => {
    return string
      .split('')
      .reverse()
      .join('');
  });
```

src/lib/booleans.js
```diff
@@ -0,0 +1,118 @@
...
+
+function none(a, b) {
+  return !a && !b;
+}
+
+function one(a, b) {
+  if (!a && !b) {
+    return false;
+  }
+  return !a || !b;
```
StuartCowley, line 21: You can rewrite this function so it will return false based on one condition, and return true if not:
```
  if (!a && !b) {
    return false;
  }
  return true;
```
```
return (!a && !b) ? false : true;
```
```
  return !(!a && !b);
```




```
  if (!a && !b) {
    return false;
  }
  return !a || !b;
```

src/lib/booleans.js
```diff
@@ -0,0 +1,118 @@
...
+  if (a === '') {
+    return false;
+  }
+  if (!a === !NaN) {
+    return false;
+  }
+  if (a == null) {
+    return false;
+  }
+  return true;
```
StuartCowley, line 36: The logic here is fine, but it can be condensed a lot using two NOT operators (!!) together to create a double negation. What this does is to first evaluate the value to a boolean that will be the opposite of what it actually is, eg passing in a string (which is a truthy value) will be made `false` by one of the ! operators, then the second ! will flip that to `true`, and so evaluating the string as being truthy. Nice trick that gets used more often than you might think!
```
  return !!a
```

src/lib/booleans.js
```diff
@@ -0,0 +1,118 @@
...
+}
+
+function isEqual(a, b) {
+  if (a === b) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 43: You can update a lot of these functions to be more concise by simply returning the evaluation of the condition, for example here:
```
return (a === b)
```

src/lib/booleans.js
```diff
@@ -0,0 +1,118 @@
...
+}
+
+function startsWith(char, string) {
+  if (string.startsWith(char) === true) {
+    return true;
+  }
+  return false;
+}
```
StuartCowley, line 86: Recursion! Very nice idea here, though you should be extremely careful with this. In this case you could rewrite the function to something like this:
```
  return string[0] === char;
```

src/lib/booleans.js
```diff
@@ -0,0 +1,118 @@
...
+
+function containsVowels(string) {
+  if (string.match('a|A|e|E|i|I|o|O|u|U')) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 92: So this feels like a good chance to introduce regex to you!

```
someString.match(/[aeiou]/i)
```
```
```
```
```
Way more concise.

src/lib/booleans.js
```diff
@@ -0,0 +1,118 @@
...
+}
+
+function isLowerCase(string) {
+  if (string === 'Abc') {
+    return false;
+  }
+  return true;
```
StuartCowley, line 99: This will not work as it is. You shgould update this and include in your test suite:
```
  return string.toLowerCase() === string;
```

src/lib/objects.js
```diff
@@ -0,0 +1,87 @@
+const createPerson = (name, age) => {
+  const person = {};
+  person.name = name;
+  person.age = age;
+  return person;
```
StuartCowley, line 5: This could be simplified to just:
```
  return {
    name,
    age,
  };
```

src/lib/objects.js
```diff
@@ -0,0 +1,87 @@
...
+    return object.age;
+  }
+  return object.name;
+};
+
+const hasProperty = (property, object) => {
+  if (object.hasOwnProperty(property)) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 23: When you are looking for return a boolean as you are here, you could use the logic you have here but with a bit of refactoring. If you were to first reduce this to a ternary:
```
```
```
  return !!object[property];
```
Much cleaner :)

src/lib/objects.js
```diff
@@ -0,0 +1,87 @@
...
+
+const getName = object => {
+  return object.name;
+};
+
+const getProperty = (property, object) => {
+  if (property === 'age') {
+    return object.age;
+  }
+  return object.name;
```
StuartCowley, line 16: This is not a good way of doing this. It passes the tests now, but what would happen if you added more properties to the object or changed one of their names? It would no longer work. Instead, you should abstract this down another layer:
```
  return object[property];
```

src/lib/objects.js
```diff
@@ -0,0 +1,87 @@
...
+    return true;
+  }
+  return false;
+};
+
+const isOver65 = person => {
+  if (person.age > 65) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 30: Can shorten this as described earlier:
```
  return person.age > 65;
```

src/lib/objects.js
```diff
@@ -0,0 +1,87 @@
...
+};
+
+const getAges = people => {
+  function getAges(input, field) {
+    var output = [];
+    for (var i = 0; i < input.length; i++) output.push(input[i][field]);
+    return output;
+  }
+  var result = getAges(people, 'age');
+  return result;
```
StuartCowley, line 40: This could be re-written. First thing to point out is that you are creating a function with the same name as the parent, which is bad practice. Also, you can make use of built in methods like filter and map to the same job for you  a bit more concisely:
```
  return people.filter(person => person.name).map(person => person.age);
```

src/lib/objects.js
```diff
@@ -0,0 +1,87 @@
...
+        'Hi ' +
+        person +
+        ', my name is ' +
+        this.name +
+        ' and I am ' +
+        this.age +
+        '!'
+      );
+    }
+  };
```
StuartCowley, line 73: The layout here is not great. You could re-write this to use string interpolation for the return string, and the name and age properties could be written using shorthand:
```
  return {
    name,
    age,
    introduce(person) {
      return `Hi ${person}, my name is ${this.name} and I am ${this.age}!`;
    },
  };
```

src/lib/strings.js
```diff
@@ -0,0 +1,32 @@
...
+  return `Hello, ${string}!`;
+}
+
+function uppercase(abcdefghi) {
+  return abcdefghi.toUpperCase();
+}
+
+function lowercase(ABCDEFGHI) {
+  return ABCDEFGHI.toLowerCase();
+}
```
StuartCowley, line 11: These variable names are not great! You should name them based on whwt they represent, in this case a string:
```
function uppercase(string) {
  return string.toUpperCase();
}

function lowercase(string) {
  return string.toLowerCase();
}
```

StuartCowley (commented):
Good stuff here! Ive left a few suggestions throughout for fixing some small errors, removing eslint exceptions that arent needed, some refactoring to make things a bit cleaner, and general formatting. Theres some nice ideas here and its good to see you trying new things! Keep it up, theres some solid work here

PR #11, opened 2023-04-30
Jose wills code review

StuartCowley:


README.md
```diff
@@ -1,38 +1,19 @@
...
+* Send a HTTP response with a custom status code and body using Express.
+* Explain the difference between an end-to-end (E2E) test and a unit test.
+
+## Concepts
+
+* Web Servers and Web API's.
+* HTTP Requests and Responses.
+* Routing.
+* Controller functions.
+* Middleware.
```
StuartCowley, line 19: It would be great if you could include a table of the routes this uses here, for example

```
## Endpoints
### Strings

| HTTP method | Method      | Endpoint | Description |
| ----------- | ----------- |----------- | ----- |
| GET         | sayHello(string) | /strings/hello/:string | Returns the string "Hello, " concatenated with the given string. |
| GET         | uppercase(string) | /strings/upper/:string | Returns the given string converted to all uppercase. |
```
You could create a new table for each grouping like this, adding whatever info you like. Have a look at [this guide](https://www.markdownguide.org/extended-syntax/) for more info on the formatting options possible

README.md
```diff
@@ -1,38 +1,19 @@
...
-```
-
-## Running tests
-
-You can run tests with the `npm test` command.
-
-## Remember: READMEs are important!
-
-When you're finished, be sure to rewrite this one so it's more specific to your project.
+## Learning objectives
```
StuartCowley, line 3: A description of what your code actually does would be good here. While the learning objectives are useful for you, they arent a massive help to anyone who may want to use your code as theres no explicit description of what it actually does

Also some installation instructions would be great, they are a standard feature of all good READMEs

src/app.js
```diff
@@ -1,5 +1,202 @@
...
+  arrayToCSVString,
+  addToArray2,
+  elementsStartingWithAVowel,
+  removeNthElement,
+} = require('./lib/arrays');
 
 const app = express();
 
+app.use(express.json());
+
```
StuartCowley, line 16: For long files like this, you could break up the large block of code with heading comments like:
```

// String methods

```

src/lib/arrays.js
```diff
@@ -0,0 +1,100 @@
+const getNthElement = (index, array) => {
+  index %= array.length;
+  return array[index];
```
StuartCowley, line 3: Did you notice 'Assignment to function parameter' eslint error in your IDE? This is not a good thing to do as it can lead to unexpected behaviours and bugs in larger functions where the value coming into the function is no longer what its expected to be, so you should always avoid doing it.
In this case you can do this using the modulus operator, which is similar to division except that instead of returning a decimal when dividing any remainder, it gives a whole number (EG 5 % 3 = 2). This means you dont have to create any variables at all:

```
```

src/lib/arrays.js
```diff
@@ -0,0 +1,100 @@
...
+    .join('')
+    .split(',')
+    .reverse();
+};
+
+const onlyEven = numbers => {
+  function even(n) {
+    return !(n % 2);
+  }
+  return numbers.filter(even);
```
StuartCowley, line 53: Theres no need to write a separate function like this, you can do it inline:
```
  return numbers.filter(number => number % 2 !== 0);
```

src/lib/arrays.js
```diff
@@ -0,0 +1,100 @@
...
+};
+
+const removeNthElement2 = (index, array) => {
+  const firstArray = array.slice(0, index);
+  const secondArray = array.slice(index + 1, array[-1]);
+  return firstArray.concat(secondArray);
```
StuartCowley, line 59: You could shorten this a bit:
```
  return firstArray.concat(array.slice(index + 1, array[-1]));
```

src/lib/arrays.js
```diff
@@ -0,0 +1,100 @@
...
+};
+
+const elementsStartingWithAVowel = strings => {
+  const vowelWords = strings.filter(word => /^[aeiou]/i.test(word));
+  return vowelWords;
```
StuartCowley, line 64: No need to create the extra variable to just return it on the next line, can just return the result of the filter method directly:
```
  return strings.filter(word => /^[aeiou]/i.test(word));
```

src/lib/booleans.js
```diff
@@ -0,0 +1,79 @@
...
+}
+
+function isEven(a) {
+  return !(a % 2);
+}
+
+function isSquare(a) {
+  const squareRoot = Math.sqrt(a);
+  const isPerfectSquare = Math.ceil(squareRoot) === Math.floor(squareRoot);
+  return isPerfectSquare;
```
StuartCowley, line 48: Can shorten this a bit:
```
  const squareRoot = Math.sqrt(a);
  return Math.ceil(squareRoot) === Math.floor(squareRoot);
```

src/lib/objects.js
```diff
@@ -0,0 +1,74 @@
...
+const getName = object => {
+  return object.name;
+};
+
+const getProperty = (property, object) => {
+  return object[property];
+};
+
+const hasProperty = (property, object) => {
+  return object.hasOwnProperty(property);
```
StuartCowley, line 18: You can fix the eslint error being thrown here by using the array style notation for accessing the properties of the object as follows:
```
  return object[property];
```

This gets rid of the error, but it doesnt return the value we want.

When you are looking for return a boolean as you are here, we could use a ternary:
```
```
While this works, it could be reduced further by using a double negation:
```
  return !!object[property];
```
Much cleaner :)

src/lib/objects.js
```diff
@@ -0,0 +1,74 @@
...
+const isOver65 = person => {
+  return person.age > 65;
+};
+
+const getAges = people => {
+  const ages = [];
+  for (let i = 0; i < people.length; i++) {
+    ages.push(people[i].age);
+  }
+  return ages;
```
StuartCowley, line 30: This could be re-written. You can make use of built in methods like filter and map to the same job for you  a bit more concisely:
```
  return people.filter(person => person.name).map(person => person.age);
```

src/lib/objects.js
```diff
@@ -0,0 +1,74 @@
...
+};
+
+const findByName = (name, people) => {
+  const foundName = people.find(person => person.name === name);
+  return foundName;
```
StuartCowley, line 35: Removing the variable assignment here:
```
  return people.find(person => person.name === name);
```

src/lib/objects.js
```diff
@@ -0,0 +1,74 @@
...
+};
+
+const findHondas = cars => {
+  const findHonda = cars.filter(car => car.manufacturer === 'Honda');
+  return findHonda;
```
StuartCowley, line 40: And also here:
```
```

StuartCowley (commented):
Good work! Everything passes and the code is cleanly formatted and well organised. It would be good to add some tests for objects at some point, and the readme could use a little bit of attention, but this aside youve got solid work here :)

PR #12, opened 2023-05-01
Jordan noble code review

StuartCowley:


src/app.js
```diff
@@ -1,5 +1,163 @@
...
+app.post('/arrays/starts-with-vowel', (req, res) => {
+  res.status(200).send({ result: elementsStartingWithAVowel(req.body.array) });
+});
+
+app.post('/arrays/remove-element', (req, res) => {
+  if (!req.query.index) {
+    res.status(200).send({ result: removeNthElement(0, req.body.array) });
+  } else {
+    res.status(200).send({ result: removeNthElement(req.query.index, req.body.array) });
+  }
```
StuartCowley, line 160: Theres a bit of unnecessary repetition here, you could update this by changing how you handle the index parameter. If its falsy, you want the index value to be 0, otherwise you want to parse it into an integer and use that. So this function can be re-written to just:

```
  const index = req.query.index ? parseInt(req.query.index, 10) : 0;
  res.status(200).json({ result: removeNthElement2(index, req.body.array) });
```

src/app.js
```diff
@@ -1,5 +1,163 @@
...
+
+app.get('/strings/lower/:id', (req, res) => {
+  res.status(200).send({ result: lowerCase(req.params.id) });
+});
+
+app.get('/strings/first-characters/:id', (req, res) => {
+  const n = req.query.length;
+  if (req.query.length) {
+    res.status(200).send({ result: firstCharacters(req.params.id, n) });
+  } else {
```
StuartCowley, line 42: This could be adjusted a little to make use of the variable a bit more, and give it a clearer name:
```
  const query = req.query.length;
  if (query) {
    res.status(200).send({ result: firstCharacters(req.params.id, query) });
  } else {
```

src/app.js
```diff
@@ -1,5 +1,163 @@
...
+  }
+});
+
+app.get('/numbers/add/:a/and/:b', (req, res) => {
+  if (Number.isNaN(Number(req.params.a)) || Number.isNaN(Number(req.params.b))) {
+    res.status(400).send({ error: 'Parameters must be valid numbers.' });
+  } else {
+    const a = Number(req.params.a);
+    const b = Number(req.params.b);
+    res.status(200).send({ result: add(a, b) });
```
StuartCowley, line 53: No need to create the variables here, this can be done in line in the return statement:
```
    res.status(200).send({ result: add(Number(req.params.a), Number(req.params.b)) });
```

src/app.js
```diff
@@ -1,5 +1,163 @@
...
+  }
+});
+
+app.get('/numbers/subtract/:a/from/:b', (req, res) => {
+  if (Number.isNaN(Number(req.params.a)) || Number.isNaN(Number(req.params.b))) {
+    res.status(400).send({ error: 'Parameters must be valid numbers.' });
+  } else {
+    const a = Number(req.params.a);
+    const b = Number(req.params.b);
+    res.status(200).send({ result: subtract(b, a) });
```
StuartCowley, line 63: As above:
```
    res.status(200).send({ result: subtract(Number(req.params.b), Number(req.params.a)) });
```

src/app.js
```diff
@@ -1,5 +1,163 @@
...
+
+app.post('/numbers/multiply', (req, res) => {
+  if (!req.body.a || !req.body.b) {
+    res.status(400).send({ error: 'Parameters "a" and "b" are required.' });
+  } else if (Number.isNaN(Number(req.body.a)) || Number.isNaN(Number(req.body.b))) {
+    res.status(400).send({ error: 'Parameters "a" and "b" must be valid numbers.' });
+  } else {
+    const a = Number(req.body.a);
+    const b = Number(req.body.b);
+    res.status(200).send({ result: multiply(a, b) });
```
StuartCowley, line 75: And again:
```
    res.status(200).send({ result: multiply(req.body.a, req.body.b) });
```

src/app.js
```diff
@@ -1,5 +1,163 @@
...
+    res.status(400).send({ error: 'Parameters "a" and "b" are required.' });
+  } else if (Number.isNaN(Number(req.body.a)) || Number.isNaN(Number(req.body.b))) {
+    res.status(400).send({ error: 'Parameters "a" and "b" must be valid numbers.' });
+  } else if (Number(req.body.a) === 0) {
+    res.status(200).send({ result: 0 });
+  } else {
+    const a = Number(req.body.a);
+    const b = Number(req.body.b);
+    res.status(200).send({ result: divide(a, b) });
+  }
```
StuartCowley, line 94: Theres a few things we could do here to make this a bit clearer reduce some of the repetition. First off we could destructure the `a` and `b` values from `req.body` and use those throughout rather than writing the whole thing out each time. This mean we can get rid of the variable declarations on line 91/92
We could also eliminate the condition on line 88, since we are already checking for that on line 80 (remember that the `===` is the strict equality operator, so you are already checking whether it is a number or not)

Putting this all together, it could be re-written like this:
```
  const { a, b } = req.body;
  if (a === 0) {
    res.status(200).send({ result: 0 });
  } else if (b === 0) {
  } else if (!a || !b) {
    res.status(400).send({ error: 'Parameters "a" and "b" are required.' });
  } else if (Number.isNaN(Number(a)) || Number.isNaN(Number(b))) {
    res.status(400).send({ error: 'Parameters "a" and "b" must be valid numbers.' });
  } else {
    res.status(200).send({ result: divide(a, b) });
  }
```

src/app.js
```diff
@@ -1,5 +1,163 @@
...
+});
+
+app.post('/numbers/remainder', (req, res) => {
```
StuartCowley, line 97: The same logic described in the function directly above this could apply be here too

src/lib/arrays.js
```diff
@@ -0,0 +1,118 @@
+const getNthElement = (index, array) => {
+  if (index > array.length - 1) {
+    return array[index - array.length];
+  }
+  return array[index];
```
```
```

src/lib/arrays.js
```diff
@@ -0,0 +1,118 @@
...
+};
+
+const removeNthElement = (index, array) => {
+  array.splice(index, 1);
+  return array;
+};
+
+const numbersToStrings = numbers => {
+  const string1 = numbers.toString();
+  return string1.split(',');
```
StuartCowley, line 32: This could be made into a one liner:
```
  return numbers.toString().split();
```

src/lib/arrays.js
```diff
@@ -0,0 +1,118 @@
...
+};
+
+const uppercaseWordsInArray = strings => {
+  const uppercasearray = [];
+  for (let n = 0; n < strings.length; n += 1) {
+    uppercasearray[n] = strings[n].toUpperCase();
+  }
+  return uppercasearray;
```
StuartCowley, line 40: You could refactor this to use the built in map method:
```
  return strings.map(string => string.toUpperCase());
```
Problem solved

src/lib/arrays.js
```diff
@@ -0,0 +1,118 @@
...
+const reverseWordsInArray = strings => {
+  const newarray = [];
+  let newstring = [];
+  for (let n = 0; n < strings.length; n += 1) {
+    newstring = strings[n].split('');
+    newstring.reverse();
+    const reversedstring = newstring.join('');
+    newarray[n] = reversedstring;
+  }
+  return newarray;
```
StuartCowley, line 52: This is another good example of when you could use the map function:
```
  return strings.map(string =>
    string
      .split('')
      .reverse()
      .join('')
  );
```

src/lib/arrays.js
```diff
@@ -0,0 +1,118 @@
...
+};
+
+const onlyEven = numbers => {
+  const evennumbers = numbers.filter(number => number % 2 === 0);
+  return evennumbers;
```
StuartCowley, line 57: You can just return this value directly, no need for the variable creation:
```
  return numbers.filter(number => number % 2 === 0);
```

src/lib/arrays.js
```diff
@@ -0,0 +1,118 @@
...
+};
+
+const removeNthElement2 = (index, array) => {
+  const firsthalf = array.slice(0, index);
+  const secondhalf = array.slice(index + 1);
+  const newarray = firsthalf.concat(secondhalf);
+  return newarray;
```
StuartCowley, line 64: As above, can simplify this a bit:
```
  return firsthalf.concat(secondhalf);
```

src/lib/arrays.js
```diff
@@ -0,0 +1,118 @@
...
+      lowercase.charAt(0) === 'i' ||
+      lowercase.charAt(0) === 'o' ||
+      lowercase.charAt(0) === 'u'
+    ) {
+      return true;
+    }
+    return false;
+  });
+
+  return newstring;
```
StuartCowley, line 82: The basic idea is here, but theres a few different ways to go about this. This has a lot of repetition, and isnt that maintainable so should be re-written. One way is to have an array of values to check for, then use a combination of the filter and includes array methods with the toLowerCase string method youre already using:

```
  const vowels = ['a', 'e', 'i', 'o', 'u'];

  return strings.filter(string => vowels.includes(string[0].toLowerCase()));
```
Much more concise and does the exact same job

src/lib/arrays.js
```diff
@@ -0,0 +1,118 @@
...
+};
+
+const removeSpaces = string => {
+  const splitstring = string.split(' ');
+  const nospaces = splitstring.join('');
+  return nospaces;
```
StuartCowley, line 88: This can be made into a one liner:
```
  return string.split(' ').join('');
```

src/lib/arrays.js
```diff
@@ -0,0 +1,118 @@
...
+};
+
+const sumNumbers = numbers => {
+  const sum = numbers.reduce((acc, currentvalue) => {
+    return acc + currentvalue;
+  }, 0);
+  return sum;
```
StuartCowley, line 95: Again, can remove the variable creation:
```
  return numbers.reduce((acc, currentvalue) => {
    return acc + currentvalue;
  }, 0);
```

src/lib/booleans.js
```diff
@@ -0,0 +1,132 @@
+const negate = a => {
+  if (a) {
+    return false;
+  }
+  return true;
```
StuartCowley, line 5: You could simply flip the input using the NOT operator and return that here:
```
  return !a;
```

src/lib/booleans.js
```diff
@@ -0,0 +1,132 @@
...
+};
+
+const both = (a, b) => {
+  if (a && b) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 12: As above, just return the result of the expression:
```
  return a && b;
```

src/lib/booleans.js
```diff
@@ -0,0 +1,132 @@
...
+};
+
+const either = (a, b) => {
+  if (a || b) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 19: And again here:
```
  return a || b;
```

src/lib/booleans.js
```diff
@@ -0,0 +1,132 @@
...
+};
+
+const none = (a, b) => {
+  if (a === false && b === false) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 26: You could rewrite this function so it will return false based on one condition, and return true if not:
```
  if (a || b) {
    return false;
  }
  return true;
```
```
```
```
  return !(a || b);
```
Much cleaner :)

src/lib/booleans.js
```diff
@@ -0,0 +1,132 @@
...
+};
+
+const one = (a, b) => {
+  if ((a && b === false) || (b && a === false)) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 33: This could be simplified as above by just returning the result of the expression:
```
  return (a && b === false) || (b && a === false);
```

src/lib/booleans.js
```diff
@@ -0,0 +1,132 @@
...
+};
+
+const truthiness = a => {
+  if (a) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 40: The logic here is fine, but it can be condensed a lot using two NOT operators (!!) together to create a double negation. What this does is to first evaluate the value to a boolean that will be the opposite of what it actually is, eg passing in a string (which is a truthy value) will be made `false` by one of the ! operators, then the second ! will flip that to `true`, and so evaluating the string as being truthy. Nice trick that gets used more often than you might think!
```
  return !!a
```

src/lib/booleans.js
```diff
@@ -0,0 +1,132 @@
...
+};
+
+const isEqual = (a, b) => {
+  if (a === b) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 47: You can just return the result of the expression here:
```
  return a === b;
```

src/lib/booleans.js
```diff
@@ -0,0 +1,132 @@
...
+};
+
+const isGreaterThan = (a, b) => {
+  if (a > b) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 54: As above:
```
  return a > b;
```

src/lib/booleans.js
```diff
@@ -0,0 +1,132 @@
...
+  }
+  return true;
+};
+
+const isEven = a => {
+  if (a % 2 === 0) {
+    return true;
+  }
+  return false;
+};
```
StuartCowley, line 76: All of these can be simplified in the same way by just returning the result of the expression

src/lib/booleans.js
```diff
@@ -0,0 +1,132 @@
...
+
+const isSquare = a => {
+  const n = Math.sqrt(a);
+  if (n * n === a) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 83: This doesnt actually work... it will always return true, since you are taking the root of a number then asserting squaring that number will be the same as what you started with. Instead you could check the root is a whole number by doing this:

```
  const n = Math.sqrt(a);
  return Math.floor(n) === Math.ceil(n);
```

src/lib/booleans.js
```diff
@@ -0,0 +1,132 @@
...
+};
+
+const startsWith = (char, string) => {
+  if (string.charAt(0) === char) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 90: As above this can be simplified:
```
  return string.charAt(0) === char;
```

src/lib/booleans.js
```diff
@@ -0,0 +1,132 @@
...
+    if (
+      lowercase.charAt(n) === 'a' ||
+      lowercase.charAt(n) === 'e' ||
+      lowercase.charAt(n) === 'i' ||
+      lowercase.charAt(n) === 'o' ||
+      lowercase.charAt(n) === 'u'
+    ) {
+      return true;
+    }
+  return false;
```
StuartCowley, line 105: So this feels like a good chance to introduce regex to you!

```
someString.match(/[aeiou]/i)
```
```
```
```
```
Way more concise.

src/lib/booleans.js
```diff
@@ -0,0 +1,132 @@
...
+};
+
+const isLowerCase = string => {
+  const firstletter = string.charAt(0);
+  if (firstletter === firstletter.toLowerCase()) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 113: This could be simplified a bit:
```
  return string.charAt(0) === string.charAt(0).toLowerCase();
```

src/lib/numbers.js
```diff
@@ -0,0 +1,61 @@
...
+const absolute = a => {
+  return Math.abs(a);
+};
+
+const quotient = (a, b) => {
+  const x = a / b;
+  if (x < 0) {
+    return Math.ceil(x);
+  }
+  return Math.floor(x);
```
StuartCowley, line 42: Theres a built in function for this: Math.trunc (see [docs here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc#:~:text=The%20Math.,by%20removing%20any%20fractional%20digits.))
```
  return Math.trunc(a / b);
```

src/lib/strings.js
```diff
@@ -0,0 +1,31 @@
+const sayHello = string => {
+  return `Hello, ${string}!`;
+};
+const upperCase = string => {
```
StuartCowley, line 4: Should always leave a clear line between each of your methods:
```
};

const upperCase = string => {
```

src/lib/objects.js
```diff
@@ -0,0 +1,75 @@
+const createPerson = (name, age) => {
+  const person = Object();
+  person.name = name;
+  person.age = age;
+  return person;
```
StuartCowley, line 5: You dont need to create an Object like this just to add some properties and return it, you can create one with the standard shorthand like so:
```
  return {
    name,
    age,
  };
```

src/lib/objects.js
```diff
@@ -0,0 +1,75 @@
...
+
+const isOver65 = person => {
+  return person.age > 65;
+};
+
+const getAges = people => {
+  const ages = people.map(person => {
+    return person.age;
+  });
+  return ages;
```
StuartCowley, line 29: You can just return the result of the map function here:
```
  return people.map(person => {
    return person.age;
  });
```

src/lib/objects.js
```diff
@@ -0,0 +1,75 @@
...
+};
+
+const findByName = (name, people) => {
+  console.log(name);
+  const find = people.find(person => person.name === name);
+  console.log(find);
+  return find;
```
StuartCowley, line 36: You can return the result of the find function here:
```
  return people.find(person => person.name === name);
```
Also, you should always remove console logs when you are done with them. In this case these wont show up anywhere when running the tests as there no tests that cover any of the methods in this file, but the point still stands

src/lib/objects.js
```diff
@@ -0,0 +1,75 @@
...
+};
+
+const findHondas = cars => {
+  const hondas = cars.filter(car => car.manufacturer === 'Honda');
+  console.log(hondas);
+  return hondas;
```
StuartCowley, line 42: As above:
```
```

src/lib/objects.js
```diff
@@ -0,0 +1,75 @@
...
+};
+
+const averageAge = people => {
+  const ages = people.map(person => person.age);
+  const sum = ages.reduce((acc, currentvalue) => {
+    return acc + currentvalue;
+  }, 0);
+  const average = sum / ages.length;
+  return average;
```
StuartCowley, line 51: As above, slight condensing:
```
  return sum / ages.length;
```

StuartCowley (commented):
Some generally solid approaches here, there are some refactoring suggestions throughout where you can make things a bit more concise or cleaner, plus a couple where the logic isnt quite right but its not showing up in the test as there are no test cases written for some of these methods. 
It also seems like the objects file wasnt quite finished as there were several console logs scattered about and it seems like youre still working on some of these.
Either way, the code is well presented and for the most part optimal! Youre doing good, keep it up

PR #13, opened 2023-05-02
Joshua kuhn code review

StuartCowley:


README.md
```diff
@@ -1,24 +1,30 @@
-# JavaScript Basics in Express
+# Express API for JavaScript Basics
 
-## Getting started
+This is an Express API that provides functionality for basic JavaScript concepts such as arrays, booleans, numbers, objects, and strings. The API has been developed to work with the previously written javascript-basics code.
+Getting started
```
StuartCowley, line 4: This should be a ## heading

README.md
```diff
@@ -31,8 +37,60 @@ src
...
-You can run tests with the `npm test` command.
+You can run tests with the command:
+
+```bash
+
+npm test
+
+```
+
+## API Endpoints
```
StuartCowley, line 48: Nice readme!
If you want to improve the look of it a bit more, you could update this section so its all in a markdown table, something like:
```
### Strings

```

src/app.js
```diff
@@ -1,5 +1,150 @@
...
+});
+
+app.get('/strings/lower/:string', (req, res) => {
+  res.status(200).send({ result: strings.lowercase(req.params.string) });
+});
+
+app.get('/strings/first-characters/:string', (req, res) => {
+  const { string } = req.params;
+  const length = +req.query.length || 1;
+  const result = strings.firstCharacters(string, length);
```
StuartCowley, line 27: You could simplify this into a clean one liner:
```
  const result = strings.firstCharacters(req.params.string, req.query.length || 1);
```
I see what you were thinking using the `+` sign for number type coercion, but in this case it isnt actually needed

src/app.js
```diff
@@ -1,5 +1,150 @@
...
+  const number = +req.params.number;
+  if (Number.isNaN(number)) {
+    res.status(400).send({ error: 'Parameter must be a number.' });
+  }
+  res.status(200).send({ result: booleans.isOdd(number) });
+});
+
+app.get('/booleans/:string/starts-with/:char', (req, res) => {
+  const { string, char } = req.params;
+  if (+char.length > 1) {
```
StuartCowley, line 116: No need for the type casting here
```
  if (char.length > 1) {
```

src/app.js
```diff
@@ -1,5 +1,150 @@
...
+  }
+  if (Number.isNaN(+a) || Number.isNaN(+b)) {
+    res.status(400).send({ error: 'Parameters must be valid numbers.' });
+  }
+  res.status(200).send({ result: numbers.remainder(a, b) });
+});
+
+app.post('/booleans/negate', (req, res) => {
+  const { value } = req.body;
+  res.status(200).send({ result: booleans.negate(value) });
```
StuartCowley, line 98: No need for the extra variable creation here:
```
  res.status(200).send({ result: booleans.negate(req.body.value) });
```

src/app.js
```diff
@@ -1,5 +1,150 @@
...
+});
+
+app.post('/booleans/truthiness', (req, res) => {
+  const { value } = req.body;
+  res.status(200).send({ result: booleans.truthiness(value) });
```
StuartCowley, line 103: As above
```
  res.status(200).send({ result: booleans.truthiness(req.body.value) });
```

src/app.js
```diff
@@ -1,5 +1,150 @@
...
+  if (+char.length > 1) {
+    res.status(400).send({ error: 'Parameter "character" must be a single character.' });
+  }
+  res.status(200).send({ result: booleans.startsWith(char, string) });
+});
+
+app.post('/arrays/element-at-index/:index', (req, res) => {
+  const { index } = req.params;
+  const { array } = req.body;
+  res.status(200).send({ result: arrays.getNthElement(index, array) });
```
StuartCowley, line 125: No need to create the variables here
```
  res.status(200).send({ result: arrays.getNthElement(req.params.index, req.body.array) });
```

src/app.js
```diff
@@ -1,5 +1,150 @@
...
+});
+
+app.post('/arrays/to-string', (req, res) => {
+  const { array } = req.body;
+  res.status(200).send({ result: arrays.arrayToCSVString(array) });
```
StuartCowley, line 130: As above
```
  res.status(200).send({ result: arrays.arrayToCSVString(req.body.array) });
```

src/app.js
```diff
@@ -1,5 +1,150 @@
...
+
+app.post('/arrays/append', (req, res) => {
+  const { value, array } = req.body;
+  arrays.addToArray(value, array);
+  res.status(200).send({ result: array });
+});
+
+app.post('/arrays/starts-with-vowel', (req, res) => {
+  const { array } = req.body;
+  res.status(200).send({ result: arrays.elementsStartingWithAVowel(array) });
```
StuartCowley, line 141: As above
```
  res.status(200).send({ result: arrays.elementsStartingWithAVowel(req.body.array) });
```

src/app.js
```diff
@@ -1,5 +1,150 @@
...
+});
+
+app.post('/arrays/remove-element', (req, res) => {
+  const { index } = req.query;
+  const { array } = req.body;
+  res.status(200).send({ result: arrays.removeNthElement2(index, array) });
```
StuartCowley, line 147: Once more as above
```
  res.status(200).send({ result: arrays.removeNthElement2(req.query.index, req.body.array) });
```

src/lib/arrays.js
```diff
@@ -0,0 +1,88 @@
+const getNthElement = (index, array) => {
+  if (index >= array.length) {
+    return array[index - array.length];
+  }
+  return array[index];
```
```
```

src/lib/booleans.js
```diff
@@ -0,0 +1,78 @@
...
+  return Math.sqrt(a) % 1 === 0;
+};
+
+const startsWith = (char, string) => {
+  return string.charAt(0) === char;
+};
+
+const containsVowels = (string) => {
+  const vowels = /[aeiou]/i;
+  return vowels.test(string);
```
StuartCowley, line 55: No need to create the extra variable here
```
  return /[aeiou]/i.test(string);
```

src/lib/objects.js
```diff
@@ -0,0 +1,60 @@
...
+  return cars.filter(a => a.manufacturer === 'Honda');
+};
+
+const averageAge = people => {
+  const sum = people.reduce(
+    (accumulative, current) => accumulative + current.age,
+    0
+  );
+  const avg = sum / people.length;
+  return Math.round(avg);
```
StuartCowley, line 39: Slight shortening here:
```
  return Math.round(sum / people.length);
```

src/lib/objects.js
```diff
@@ -0,0 +1,60 @@
...
+const getAges = people => {
+  return people.map(person => person.age);
+};
+
+const findByName = (name, people) => {
+  return people.find(person => person.name === name);
+};
+
+const findHondas = cars => {
+  return cars.filter(a => a.manufacturer === 'Honda');
```
StuartCowley, line 30: Id always suggest using more descriptive variable names, even if it is just in a filter method like this. Its just a good habit to get into
```
```

StuartCowley (commented):
Very good work here, have just left some suggestions mostly around minor readability improvements and the readme. Would be good to see some tests included for the objects at some point in the future, but as this stands all the tests pass and its nice and readable. Keep it up!

PR #14, opened 2023-05-02
Benj speight code review

StuartCowley:


src/app.js
```diff
@@ -2,4 +2,151 @@ const express = require('express');
...
+  }
+  if (Number.isNaN(+a) || Number.isNaN(+b)) {
+    res.status(400).json({ error: 'Parameters must be valid numbers.' });
+  }
+  res.status(200).json({ result: numbers.remainder(a, b) });
+});
+
+app.post('/booleans/negate', (req, res) => {
+  const { value } = req.body;
+  res.status(200).json({ result: booleans.negate(value) });
```
StuartCowley, line 99: This could easily be done in one line:
```
  res.status(200).json({ result: booleans.negate(req.body.value) });
```

src/app.js
```diff
@@ -2,4 +2,151 @@ const express = require('express');
...
+});
+
+app.post('/booleans/truthiness', (req, res) => {
+  const { value } = req.body;
+  res.status(200).json({ result: booleans.truthiness(value) });
```
StuartCowley, line 104: As could this:
```
  res.status(200).json({ result: booleans.truthiness(req.body.value) });
```

src/app.js
```diff
@@ -2,4 +2,151 @@ const express = require('express');
...
+  const { char, string } = req.params;
+  if (char.length > 1) {
+    res.status(400).json({ error: 'Parameter "character" must be a single character.' });
+  } else res.status(200).json({ result: booleans.startsWith(char, string) });
+});
+
+app.post('/arrays/element-at-index/:i', (req, res) => {
+  const { i } = req.params;
+  const { array } = req.body;
+  res.status(200).json({ result: arrays.getNthElement(i, array) });
```
StuartCowley, line 125: This can be put into an easily readable one liner too:
```
  res.status(200).json({ result: arrays.getNthElement(req.params.i, req.body.array) });
```

src/app.js
```diff
@@ -2,4 +2,151 @@ const express = require('express');
...
+});
+
+app.post('/arrays/to-string', (req, res) => {
+  const { array } = req.body;
+  res.status(200).json({ result: arrays.arrayToCSVString(array) });
```
StuartCowley, line 130: Same here:
```
  res.status(200).json({ result: arrays.arrayToCSVString(req.body.array) });
```

src/app.js
```diff
@@ -2,4 +2,151 @@ const express = require('express');
...
+});
+
+app.post('/arrays/append', (req, res) => {
+  const { value, array } = req.body;
+  res.status(200).json({ result: arrays.addToArray2(value, array) });
+});
+
+app.post('/arrays/starts-with-vowel', (req, res) => {
+  const { array } = req.body;
+  res.status(200).json({ result: arrays.elementsStartingWithAVowel(array) });
```
StuartCowley, line 140: And again here:
```
  res.status(200).json({ result: arrays.elementsStartingWithAVowel(req.body.array) });
```

src/lib/arrays.js
```diff
@@ -0,0 +1,98 @@
+const getNthElement = (index, array) => {
+  const { length } = array;
+  const adjustedIndex = index % length;
+  return array[adjustedIndex];
```
StuartCowley, line 4: You can simplify this quite a bit using the modulus operator, which is similar to division except that instead of returning a decimal when dividing any remainder, it gives a whole number (EG 5 % 3 = 2). 
```
```

src/lib/arrays.js
```diff
@@ -0,0 +1,98 @@
...
+  array.splice(index, 1);
+};
+
+const numbersToStrings = numbers => {
+  return numbers.map(num => num.toString());
+};
+
+const uppercaseWordsInArray = strings => {
+  // eslint-disable-next-line no-shadow
+  return strings.map(strings => strings.toUpperCase());
```
StuartCowley, line 33: The eslint error youre getting here can be avoided without using this exception, its to do with syntax you're using for the `map` function. Hovering over the text in your IDE will show a description of the error, in this case `'strings' is already declared in the upper scope`. This is telling you are are trying to use a value for a temporary variable that already exists, which is a bad practice.
The parameter given to the callback function in functions like `map` or `filter` can be anything you like, but it makes sense to use something semantic that makes logical sense. In this case it could be swapped out for 'string':

```
  return strings.map(string => string.toUpperCase());
```
Problem solved

src/lib/arrays.js
```diff
@@ -0,0 +1,98 @@
...
+};
+
+const reverseWordsInArray = strings => {
+  // eslint-disable-next-line no-shadow
+  return strings.map(strings =>
+    strings
+      .split('')
+      .reverse()
+      .join('')
+  );
```
StuartCowley, line 43: Same issue as above:
```
  return strings.map(string =>
    string
      .split('')
      .reverse()
      .join('')
  );
```

src/lib/arrays.js
```diff
@@ -0,0 +1,98 @@
...
+};
+
+const removeNthElement2 = (index, array) => {
+  return array.filter((_, i) => i !== index);
+};
+
+const elementsStartingWithAVowel = strings => {
+  const vowels = ['a', 'e', 'i', 'o', 'u'];
+  // eslint-disable-next-line no-shadow
+  return strings.filter(strings => vowels.includes(strings[0].toLowerCase()));
```
StuartCowley, line 57: As before:
```
  return strings.filter(string => vowels.includes(string[0].toLowerCase()));
```

src/lib/arrays.js
```diff
@@ -0,0 +1,98 @@
...
+      .join('')
+  );
+};
+
+const onlyEven = numbers => {
+  return numbers.filter(number => number % 2 === 0);
+};
+
+const removeNthElement2 = (index, array) => {
+  return array.filter((_, i) => i !== index);
```
StuartCowley, line 51: Very nice!

src/lib/arrays.js
```diff
@@ -0,0 +1,98 @@
...
+    const lastA = a.charAt(a.length - 1);
+    const lastB = b.charAt(b.length - 1);
+    if (lastA < lastB) {
+      return -1;
+    }
+    if (lastA > lastB) {
+      return 1;
+    }
+    return 0;
+  });
```
StuartCowley, line 79: This could be made a bit more concise with a bit of refactoring:
```
  return strings.sort((a, b) => a.charAt(a.length - 1) - b.charAt(b.length - 1));
```

src/lib/booleans.js
```diff
@@ -0,0 +1,97 @@
...
+function isLessThanOrEqualTo(a, b) {
+  return (a <= b);
+};
+
+function isOdd(a) {
+  if (a % 2==0){
+    return false;}
+ else {
+    return true;
+ }
```
StuartCowley, line 42: This could be simplified as above by just returning the result of the expression:
```
  return a % 2 === 0;
```

src/lib/booleans.js
```diff
@@ -0,0 +1,97 @@
...
+    return true;
+} else {
+    return false;
+}
+};
+
+function startsWith(char, string) {
+  if (string.startsWith(char)) {return true;}
+  else {return false;}
+};
```
StuartCowley, line 64: As with the isOdd function these could all be simplified

src/lib/booleans.js
```diff
@@ -0,0 +1,97 @@
...
+
+function containsVowels(string) {
+  string = string.toLowerCase();
+  for (var i = 0; i < string.length; i++) {
+    if (string[i] == 'a' || string[i] == 'e' || string[i] == 'i' || string[i] == 'o' || string[i] == 'u') {
+      return true;
+    }
+  }
+  return false;
```
StuartCowley, line 73: So this feels like a good chance to talk a bit about regex!

```
someString.match(/[aeiou]/i)
```
```
```
```
```
Way more concise.

src/lib/booleans.js
```diff
@@ -0,0 +1,97 @@
...
+};
+
+function isLowerCase(string) {
+  return string === string.toLowerCase();
+
+};
```
StuartCowley, line 79: Dont need this extra blank line here:
```
  return string === string.toLowerCase();
};
```

src/lib/numbers.js
```diff
@@ -0,0 +1,57 @@
...
+const multiply = (a, b) => {
+  return a * b;
+};
+
+const divide = (a, b) => {
+  return a / b;
+};
+
+const power = (a, b) => {
+  return Math.pow(a, b);
```
StuartCowley, line 18: You have an eslint error on this line. Remember if you hover over the code with red lines below it VSCode shows some info on what the issue is and how to resolve it, in this case it tells you to use the exponentiation operator instead:
```
  return a ** b;
```

src/lib/numbers.js
```diff
@@ -0,0 +1,57 @@
...
+const roundDown = a => {
+  return Math.floor(a);
+};
+
+const absolute = a => {
+  return Math.abs(a);
+};
+
+const quotient = (a, b) => {
+  return ~~(a / b);
```
StuartCowley, line 38: Eslint is throwing an error here. As it happens theres a built in function for this: Math.trunc (see [docs here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc#:~:text=The%20Math.,by%20removing%20any%20fractional%20digits.))
```
  return Math.trunc(a / b);
```

src/lib/objects.js
```diff
@@ -0,0 +1,62 @@
+const createPerson = (name, age) => {
+  return {
+    name: name,
+    age: age
```
StuartCowley, line 4: Eslint is complaining here as you can use the shorthand form for this:
```
    name,
    age,
```

src/lib/objects.js
```diff
@@ -0,0 +1,62 @@
...
+  return totalAge / people.length;
+};
+
+const createTalkingPerson = (name, age) => {
+  return {
+    name: name,
+    age: age,
+    introduce: (friendName) => {
+      return `Hi ${friendName}, my name is ${name} and I am ${age}!`;
+    }
```
StuartCowley, line 47: As above you can use the shorthand for lines 43/44, and if you remove the brackets around the argument for the function it will stop complaining there too:
```
    name,
    age,
    introduce: friendName => {
      return `Hi ${friendName}, my name is ${name} and I am ${age}!`;
    },
```

StuartCowley (commented):
Theres a good start here, theres a few of the methods which arent covered anywhere which would be really nice to include for a bit of extra practice.
Also, your readme would be a great place to note what these methods are, and what endpoints test them. You could add a table in markdown format to the readme such as:
```
### Strings

```

What you have here is good, some nice approaches. Have added a few things to improve the flow here and there and tidy things up a bit, if you only have time to either add more methods to your routes or to update the readme, I would do the readme as it will be easier to come back to this later on and see where you were up to should you want to round it out

PR #15, opened 2023-05-03
Camilla priest stephens code review

StuartCowley:


README.md
```diff
@@ -1,38 +1,37 @@
 # JavaScript Basics in Express
 
-## Getting started
+This repository contains a series of controller functions I wrote using Express to create API endpoints for an application. It uses functions from my JavaScript Basics code, which are called when each controller function runs.
```
StuartCowley, line 3: Nice readme! Id suggest you add a link to your linkedIn profile somewhere, Nice readme!
If you want to improve the look of it a bit more, also it would be awesome if you could add some details about the routes and what they do here, a nice way to do this would be in a table, something like:
```
### Strings

```

src/app.js
```diff
@@ -1,5 +1,165 @@
...
+  res.status(200).json({ result: uppercase(req.params.string) });
+});
+
+app.get('/strings/lower/:string', (req, res) => {
+  res.status(200).json({ result: lowercase(req.params.string) });
+});
+
+app.get('/strings/first-characters/:string', (req, res) => {
+  const n = req.query.length;
+  if (req.query.length) {
```
StuartCowley, line 43: Why not make use of your variable here?
```
  const n = req.query.length;
  if (n) {
```

src/app.js
```diff
@@ -1,5 +1,165 @@
...
+app.post('/arrays/starts-with-vowel', (req, res) => {
+  res.status(200).json({ result: elementsStartingWithAVowel(req.body.array) });
+});
+
+app.post('/arrays/remove-element', (req, res) => {
+  if (req.query.index) {
+    res.status(200).json({ result: removeNthElement2(req.query.index, req.body.array) });
+  } else {
+    res.status(200).json({ result: removeNthElement2(0, req.body.array) });
+  }
```
StuartCowley, line 162: While this works, theres a bit of repetition here that isnt needed. You could simplify this with a ternary operator like this:
```
  res
    .status(200)
    .json({ result: removeNthElement2(req.query.index ? req.query.index : 0, req.body.array) });
```

Going a bit further, you could also use a [null coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) and phrase the same things thing like this:
```
  res
    .status(200)
    .json({ result: removeNthElement2(req.query.index ?? 0, req.body.array) });
```

src/lib/booleans.js
```diff
@@ -0,0 +1,80 @@
...
+
+function startsWith(char, string) {
+  return string.startsWith(char);
+};
+
+function containsVowels(string) {
+  if (string.match(/[aeiou]/gi)) {
+    return true;
+  }
+    return false;
```
StuartCowley, line 57: You can tweak this to make it a one liner. The `.match` function in the way youre using it here will either return the first instance it finds, or `null` if it finds no matches, so we can turn this into a boolean using the double negation trick (!!) that you are using for the truthiness function above:
```
  return !!string.match(/[a|e|i|o|u|]/i, 0);
```
Here if the result is a match, we dont want to return the value but a `true` boolean, so we can use a single ! to turn that truthy value false, then another to flip it back to true as we want.
If there is no match it will return `null` (a falsy value but _not strictly equal to false_), by the same logic we can flip it to truthy and back to falsy and give us the result we want :)

src/lib/numbers.js
```diff
@@ -0,0 +1,57 @@
...
+function roundDown (a) {
+  return Math.floor(a);
+}
+
+function absolute (a) {
+  return Math.abs(a);
+}
+
+function quotient (a, b) {
+  return Math.trunc(a/b);
```
StuartCowley, line 38: Slight formatting tweak:
```
  return Math.trunc(a / b);
```

StuartCowley (commented):
This is really good work! Everything is clean and formatted well, the tests all pass and the readme gives a good introduction into the repo and what it does. Have left some comments on what you could add to this, as well as some small tweaks throughout the code.

If you wanted to get some more practice you could always try adding some tests for the Object methods, plus a few of the others which arent getting used right now, but thats entirely up to you of course. Generally though this is great, keep it up!

PR #16, opened 2023-05-22
Matt Mullins code review

StuartCowley:


__tests__/strings.test.js
```diff
@@ -14,50 +14,61 @@ describe('/strings', () => {
...
+      .get('/strings/first-characters/sd32fg45')
+      .query({ length: 4 })
+      .then(res => {
+        expect(res.status).toEqual(200);
+        expect(res.body).toEqual({ result: 'sd32' });
+        done();
+      });
   });
 });
+// });
```
StuartCowley, line 74: This should be removed, theres no reason to leave a commented out brace like this in here

src/app.js
```diff
@@ -1,5 +1,169 @@
...
+});
+
+app.get('/strings/first-characters/:string', (req, res) => {
+  res.status(200).json({ result: firstCharacter(req.params.string) });
+});
+
+// doesnt work due to req.query not used, not sure what to do
+app.get('/strings/first-characters/:string', (req, res) => {
+  res.status(200).json({ result: firstCharacters(req.params.string) });
+});
```
StuartCowley, line 47: Okay, so this is failing because you are using the same endpoint twice (on lines 40 and 45), and there is no conditional logic to define which of the methods (`firstCharacter` or `firstCharacters`) should be used. At present, this endpoint is always handled by the first of these 2, so even if there are values being passed in the query they are not boing to be read because express has already handled these requests in the first one (and failed as there was no way for it to grab the values being passed in as a query)
One way to handle this is to use the same endpoint, but have some conditional logic to check whether or not there is a query present to determine which method to call. Something like this would work:
```
  const length = req.query.length
  if (length) {
    res.status(200).json({ result: firstCharacters(req.params.string, length) });
  } else {
  }
});
```
That handles both eventualities in one block, and the tests will now pass

__tests__/strings.test.js
```diff
@@ -14,50 +14,61 @@ describe('/strings', () => {
...
+    request(app)
+      .get('/strings/first-characters/hello')
+      .then(res => {
+        expect(res.status).toEqual(200);
+        expect(res.body).toEqual({ result: 'h' });
+        done();
+      });
+  });
+
+  xit('returns the first n character of the string when passed a query parameter', done => {
```
StuartCowley, line 63: This shouldnt be skipped!

src/lib/arrays.js
```diff
@@ -0,0 +1,97 @@
...
+};
+
+const onlyEven = numbers => {
+  // your code here
+  return numbers.filter(numbers => numbers % 2 === 0);
+};
+
+const removeNthElement2 = (index, array) => {
+  // your code here
+  return array.slice(0, index).concat(array.slice(index + 1));
```
StuartCowley, line 54: This logic doesnt work. You should replace it with something like a filter method, which does what you want it to right out of the box:
```
  return array.filter(element => element !== array[index]);
```

StuartCowley (commented):
Theres a few things to look at here. Some of the tests are failing, one of which I saw your comment about so have added a solution to get you through it. A major point is that the suggestions from the last code review for the js basics have not been applied here, so I havent left any comments regarding formatting, let/const declarations etc as I would just be repeating what youve already been told, instead have just focused on getting the tests all to pass and get the bugs sorted out.

I know its early days but its worth mentioning that you shouldnt be added all your work together in one massive commit, but instead be gradually adding small and concise commits as you go along. Not a major problem right now, but this is the standard industry approach to handling version control so its worth getting some practice in from the get go

Theres enough here to get everything working again for you, so I would adivse you put the suggestions from the last review into action and add these too, that way you get the best of both worlds

PR #17, opened 2023-05-22
Dave bratby code review

StuartCowley:


__tests__/strings.test.js
```diff
@@ -3,7 +3,7 @@ const app = require('../src/app');
 
 describe('/strings', () => {
   describe('GET /hello/{string}', () => {
-    it('returns "Hello world!" when passed "world"', done => {
+    xit('returns "Hello world!" when passed "world"', done => {
```
StuartCowley, line 6: You need to change all these `xit` to `it` to run these too. As it turns out theres only one of these that isnt actually working as submitted so theres no point in skipping all these

src/lib/booleans.js
```diff
@@ -0,0 +1,121 @@
...
+    return true;
+  }
+  return false;
+}
+
+function none(a, b) {
+  if (!a && !b) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 26: You can rewrite this function so it will return false based on one condition, and return true if not:
```
  if (a || b) {
    return false;
  }
  return true;
```
```
```
```
  return !(a || b);
```

src/lib/booleans.js
```diff
@@ -0,0 +1,121 @@
+function negate(a) {
+  if (a === true) {
+    return false;
+  }
+  return true;
```
StuartCowley, line 5: You can simplify this a fair bit:
```
  return !a;
```

src/lib/booleans.js
```diff
@@ -0,0 +1,121 @@
...
+}
+
+function both(a, b) {
+  if (a && b === true) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 12: You can rewrite this function so it will return true based on a single condition:
```
  if (a && b) {
    return true;
  }
  return false;
```
```
return (a && b) ? true : false;
```
```
  return (a && b);
```

src/lib/booleans.js
```diff
@@ -0,0 +1,121 @@
...
+}
+
+function either(a, b) {
+  if (a || b === true) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 19: This could be reduced again:
```
  return a || b;
```

src/lib/booleans.js
```diff
@@ -0,0 +1,121 @@
...
+}
+
+function one(a, b) {
+  if ((a && !b) || (!a && b)) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 33: This could be simplified as above by just returning the result of the expression:
```
  return (a && b === false) || (b && a === false);
```

src/lib/booleans.js
```diff
@@ -0,0 +1,121 @@
...
+
+function truthiness(a) {
+  return Boolean(a);
+}
+
+function isEqual(a, b) {
+  if (a === b) {
+    return true;
+  }
+  return false;
```
StuartCowley, line 44: You can update a lot of these functions to be more concise by simply returning the evaluation of the condition, for example here:
```
return (a === b)
```

src/lib/booleans.js
```diff
@@ -0,0 +1,121 @@
...
+}
+
+function isEven(a) {
+  if (a % 2 === 0) {
+    return true;
+  }
+  return false;
+}
```
StuartCowley, line 73: All of these could be simplified by returning just the result of the expression. Rather than adding a solution for all of these as I think youve already had these functions reviewed in the original js basics repo (which ideally would have been updated before stating this one) Ill defer to the comments on that one and just focus on express related issues as theres not much point in me repeating myself

If this isnt the case and you never got these then do please let me know if you have any questions about how to solve these things as its important you understand the basics

src/lib/objects.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+};
+
+const getName = object => {
+  return object.name
+};
+
+const getProperty = (property, object) => {
+  return object[property]
+  // use square bracket notation to access a property using its variable name
```
StuartCowley, line 14: Dont need the comment here

src/lib/arrays.js
```diff
@@ -0,0 +1,104 @@
...
+  // create a new array to hold the result
+  const result = [];
+  // loop through the current array
+  for (let i = 0; i < array.length; i++) {
+    // skip the nth element here
+    if (i !== index) {
+      result.push(array[i]);
+    }
+  }
+  // return the result array with the element removed
```
StuartCowley, line 56: Dont need these comments here, same applies to the comment below too

src/app.js
```diff
@@ -1,5 +1,163 @@
...
+  const firstChar = string ? string.charAt(0) : 'h';
+  res.status(200).json({ result: firstChar });
+});
+
+app.get('/strings/first-characters/:string', (req, res) => {
+  const { string } = req.params;
+  const { length } = req.query;
+  const firstChars = string.slice(0, length);
+  res.json({ result: firstChars });
+});
```
StuartCowley, line 41: Okay, so this is failing for a few reasons. The first one only passes because youve hard coded the result of the test in so its a false positive. The second one isnt using the method you wrote in your js basics work either, so it should be updated.
You are also using the same endpoint twice (on lines 30 and 46), and there is no conditional logic to define which of the methods (`firstCharacter` or `firstCharacters`) should be used. At present, this endpoint is always handled by the first of these 2, so even if there are values being passed in the query they are not boing to be read because express has already handled these requests in the first one (and failed as there was no way for it to grab the values being passed in as a query)
One way to handle this is to use the same endpoint, but have some conditional logic to check whether or not there is a query present to determine which method to call. Something like this would work:
```
  const { length } = req.query;
  if (length) {
    res.status(200).json({ result: firstCharacters(req.params.string, length) });
  } else {
  }
});
```
Also you will have to import the 2 functions at the top of this file for this to work

StuartCowley (commented):
Most of your tests pass, but there are a few issues with a couple of the string methods which Ive detailed on the `app.js` file. Ideally this repo would have been completed after implementing the suggestions from the initial code review of js basics, so I havent focused on suggesting fixes for the functions themselves so Im not just repeating whats already been said

If you could implement the suggestions from the initial review and the express specific stuff in this one the repo will be in a much better place
As always if you have any questions please ask in the cohort channel and one of the tutors will help you out :)

